PR #1, opened 2024-11-25
Contributing valkeyJSON module

roshkhatri:
This PR adds the implementation of valkeyJSON module based on the specifications detailed in the accepted RFC document: [ValkeyJSON Specification](https://github.com/valkey-io/valkey-rfc/blob/main/ValkeyJSON.md).

The implementation does not include JSON.MSET implementation but I am working on it as we review this. I am hoping we could add the implementation post merging this PR.

@valkey-io/core-team

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,27 @@
+name: ci
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  build-ubuntu-latest:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        server_version: ['unstable', '8.0.0']
+    steps:
+    - uses: actions/checkout@v4
+    - name: Set the server verison for python integeration tests
+      run: echo "SERVER_VERSION=${{ matrix.server_version }}" >> $GITHUB_ENV
+    - name: Set up Python
+      uses: actions/setup-python@v3
+      with:
+        python-version: '3.9'
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install -r requirements.txt
+    - name: Build and Run tests. 
+      run: ./build.sh
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,8 +1,6 @@
-name: ci
+name: CI
 
-on:
-  push:
-  pull_request:
+on: [push, pull_request]
 
 jobs:
   build-ubuntu-latest:
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,7 +26,7 @@ set(JSON_MODULE_LIB json)
 
 # Define the Valkey directories
 set(VALKEY_DOWNLOAD_DIR "${CMAKE_BINARY_DIR}/_deps/valkey-src")
-set(VALKEY_BIN_DIR "${CMAKE_BINARY_DIR}/_deps/valkey-src/src/valkey/bin")
+set(VALKEY_BIN_DIR "${CMAKE_BINARY_DIR}/_deps/valkey-src/src/valkey/src")
 
 # Download and build Valkey
 ExternalProject_Add(
@@ -34,7 +34,8 @@ ExternalProject_Add(
     GIT_REPOSITORY https://github.com/valkey-io/valkey.git  # Replace with actual URL
     GIT_TAG ${VALKEY_VERSION}
     PREFIX ${VALKEY_DOWNLOAD_DIR}
-    BUILD_COMMAND make -j
+    CONFIGURE_COMMAND ""
+    BUILD_COMMAND make distclean && make -j
     INSTALL_COMMAND ""
     BUILD_IN_SOURCE 1
 )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,166 @@
+cmake_minimum_required(VERSION 3.17)
+
+include(FetchContent)
+include(ExternalProject)
+
+# Detect the system architecture
+EXECUTE_PROCESS(
+    COMMAND uname -m
+    COMMAND tr -d '\n'
+    OUTPUT_VARIABLE ARCHITECTURE
+)
+
+if("${ARCHITECTURE}" STREQUAL "x86_64")
+    message("Building JSON for x86_64")
+elseif("${ARCHITECTURE}" STREQUAL "aarch64")
+    message("Building JSON for aarch64")
+else()
+    message(FATAL_ERROR "Unsupported architecture. JSON is only supported on x86_64 and aarch64.")
+endif()
+
+# Project definition
+project(ValkeyJSONModule VERSION 1.0 LANGUAGES C CXX)
+
+# Set the name of the JSON shared library
+set(JSON_MODULE_LIB json)
+
+# Define the Valkey directories
+set(VALKEY_DOWNLOAD_DIR "${CMAKE_BINARY_DIR}/_deps/valkey-src")
+set(VALKEY_BIN_DIR "${CMAKE_BINARY_DIR}/_deps/valkey-src/src/valkey/bin")
+
+# Download and build Valkey
+ExternalProject_Add(
+    valkey
+    GIT_REPOSITORY https://github.com/valkey-io/valkey.git  # Replace with actual URL
+    GIT_TAG ${VALKEY_VERSION}
+    PREFIX ${VALKEY_DOWNLOAD_DIR}
+    BUILD_COMMAND make -j
+    INSTALL_COMMAND ""
+    BUILD_IN_SOURCE 1
+)
+
+# Define the paths for the copied files
+set(VALKEY_INCLUDE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/src/include")
+set(VALKEY_BINARY_DEST "${CMAKE_CURRENT_SOURCE_DIR}/tst/integration/.build/binaries/${VALKEY_VERSION}")
+
+ExternalProject_Add_Step(
+    valkey
+    copy_header_files
+    COMMENT "Copying header files to include/ directory"
+    DEPENDEES download
+    DEPENDERS configure
+    COMMAND ${CMAKE_COMMAND} -E make_directory ${VALKEY_INCLUDE_DIR}
+    COMMAND ${CMAKE_COMMAND} -E copy ${VALKEY_DOWNLOAD_DIR}/src/valkey/src/valkeymodule.h ${VALKEY_INCLUDE_DIR}/valkeymodule.h
+    ALWAYS 1
+)
+
+# Copy header and binary after Valkey make
+add_custom_command(TARGET valkey
+    POST_BUILD
+    COMMAND ${CMAKE_COMMAND} -E make_directory ${VALKEY_BINARY_DEST}
+    COMMAND ${CMAKE_COMMAND} -E copy ${VALKEY_BIN_DIR}/valkey-server ${VALKEY_BINARY_DEST}/valkey-server
+    COMMENT "Copied valkeymodule.h and valkey-server to destination directories"
+)
+
+# Define valkey-bloom branch
+set(VALKEY_BLOOM_BRANCH "unstable" CACHE STRING "Valkey-bloom branch to use")
+
+# Set the download directory for Valkey-bloom
+set(VALKEY_BLOOM_DOWNLOAD_DIR "${CMAKE_CURRENT_BINARY_DIR}/_deps/valkey-bloom-src")
+
+# Download valkey-bloom
+ExternalProject_Add(
+    valkey-bloom
+    GIT_REPOSITORY https://github.com/valkey-io/valkey-bloom.git
+    GIT_TAG ${VALKEY_BLOOM_BRANCH}
+    GIT_SHALLOW TRUE
+    PREFIX "${VALKEY_BLOOM_DOWNLOAD_DIR}"
+    CONFIGURE_COMMAND ""
+    BUILD_COMMAND ""
+    INSTALL_COMMAND ""
+)
+
+# Step to copy pytest files
+ExternalProject_Add_Step(
+    valkey-bloom
+    copy_pytest_files
+    COMMENT "Copying pytest files to tst/integration directory"
+    DEPENDEES build
+    COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_CURRENT_SOURCE_DIR}/tst/integration
+    COMMAND ${CMAKE_COMMAND} -E copy_directory ${VALKEY_BLOOM_DOWNLOAD_DIR}/src/valkey-bloom/tests/valkeytests ${CMAKE_CURRENT_SOURCE_DIR}/tst/integration/valkeytests
+)
+
+# Enable instrumentation options if requested
+if("$ENV{INSTRUMENT_V2PATH}" STREQUAL "yes")
+    add_compile_definitions(INSTRUMENT_V2PATH)
+    message("Enabled INSTRUMENT_V2PATH")
+endif()
+
+# Disable Doxygen documentation generation
+set(BUILD_DOCUMENTATION OFF)
+# When CODE_COVERAGE is ON, the package is built twice, once for debug and once for release.
+# To fix the problem, disable the code coverage.
+set(CODE_COVERAGE OFF)
+
+# Fix for linking error when code coverage is enabled on ARM
+if(CODE_COVERAGE AND CMAKE_BUILD_TYPE STREQUAL "Debug")
+    add_link_options("--coverage")
+endif()
+
+# Set C & C++ standard versions
+set(CMAKE_C_STANDARD 11)
+set(CMAKE_C_STANDARD_REQUIRED True)
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED True)
+
+# Always include debug symbols and optimize the code
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -O2 -g")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O2 -g")
+
+# RapidJSON SIMD optimization
+if("${ARCHITECTURE}" STREQUAL "x86_64")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=nehalem")
+elseif("${ARCHITECTURE}" STREQUAL "aarch64")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=armv8-a")
+else()
+    message(FATAL_ERROR "Unsupported architecture. JSON is only supported on x86_64 and aarch64.")
+endif()
+
+# Additional flags for all architectures
+set(ADDITIONAL_FLAGS "-fPIC")
+
+# Compiler warning flags
+set(C_WARNING "-Wall -Werror -Wextra")
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${ADDITIONAL_FLAGS} ${C_WARNING}")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${ADDITIONAL_FLAGS} ${C_WARNING}")
+
+# Fetch RapidJSON
+FetchContent_Declare(
+    rapidjson
+    GIT_REPOSITORY https://github.com/Tencent/rapidjson.git
+    GIT_TAG 0d4517f15a8d7167ba9ae67f3f22a559ca841e3b
+)
+
+# Disable RapidJSON tests and examples
+set(RAPIDJSON_BUILD_TESTS OFF CACHE BOOL "Build rapidjson tests" FORCE)
+set(RAPIDJSON_BUILD_EXAMPLES OFF CACHE BOOL "Build rapidjson examples" FORCE)
+set(RAPIDJSON_BUILD_DOC OFF CACHE BOOL "Build rapidjson documentation" FORCE)
+
+# Make Rapidjson available
+FetchContent_MakeAvailable(rapidjson)
+
+# Add the src subdirectory for building
+add_subdirectory(src)
+
+# Add the src subdirectory for building
+add_subdirectory(tst)
+
+add_custom_target(test
+    COMMENT "Run JSON integration tests."
+    USES_TERMINAL
+    COMMAND rm -rf ${CMAKE_BINARY_DIR}/tst/integration
+    COMMAND mkdir -p ${CMAKE_BINARY_DIR}/tst/integration
+    COMMAND cp -rp ${CMAKE_SOURCE_DIR}/tst/integration/. ${CMAKE_BINARY_DIR}/tst/integration/
+    COMMAND echo "[TARGET] begin integration tests"
+    COMMAND ${CMAKE_SOURCE_DIR}/tst/integration/run.sh "test" ${CMAKE_SOURCE_DIR} 
+    COMMAND echo "[TARGET] end integration tests")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,8 +115,8 @@ set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED True)
 
 # Always include debug symbols and optimize the code
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -O2 -g")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O2 -g")
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -O3 -g -fno-omit-frame-pointer")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O3 -g -fno-omit-frame-pointer")
 
 # RapidJSON SIMD optimization
 if("${ARCHITECTURE}" STREQUAL "x86_64")
@@ -150,10 +150,8 @@ set(RAPIDJSON_BUILD_DOC OFF CACHE BOOL "Build rapidjson documentation" FORCE)
 # Make Rapidjson available
 FetchContent_MakeAvailable(rapidjson)
 
-# Add the src subdirectory for building
 add_subdirectory(src)
 
-# Add the src subdirectory for building
 add_subdirectory(tst)
 
 add_custom_target(test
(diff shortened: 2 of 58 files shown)


src/json/json_api.h
```diff
@@ -0,0 +1,70 @@
+/**
+ * JSON C API for Search Module
```
madolson: What's a search module?

hwware:
Sorry，what's the blocker for  JSON.MSET implementation? ^_^?

roshkhatri:
> Sorry，what's the blocker for JSON.MSET implementation? ^_^?

There is no blocker. its WIP. That can be added in a separate PR.

CMakeLists.txt
```diff
@@ -0,0 +1,167 @@
...
+# Define valkey-bloom branch
+set(VALKEY_BLOOM_BRANCH "unstable" CACHE STRING "Valkey-bloom branch to use")
+
+# Set the download directory for Valkey-bloom
+set(VALKEY_BLOOM_DOWNLOAD_DIR "${CMAKE_CURRENT_BINARY_DIR}/_deps/valkey-bloom-src")
+
+# Download valkey-bloom
+ExternalProject_Add(
+    valkey-bloom
+    GIT_REPOSITORY https://github.com/valkey-io/valkey-bloom.git
```
zuiderkwast, line 75: I would like to see some way to not download dependencies during the build, at least optionally.

Offline builds are common to guarantee that a project doesn't have any online dependencies that have not been verified in advance. Supply chain attacks is the main concern. License/legal problems is another.

I suggest we add some environmant variable to provide the dependencies using a local paths and CMake can use these instead of downloading them if they are provided. It can be added in the future though.

zuiderkwast (approved):
There is no way that I can review all this. 27K in one commit. :D

There will need to be maintainer team for each module that can review each other's PRs. So, at least two maintainers would be useful. Is there anyone besides you @roshkhatri?

I'd say let's merge it! Later, we can talk about beta releases and stable releases. The TSC can oversee your release plans and the work on a high level, but I don't see that TSC members have enough time to be deeply involved in these modules.

CMakeLists.txt
zuiderkwast: > I would like to see some way to not download dependencies during the build, at least optionally.
> 
> Offline builds are common to guarantee that a project doesn't have any online dependencies that have not been verified in advance. Supply chain attacks is the main concern. (License issues is another concern.)
> 
> I suggest we add some environmant variable to provide the dependencies using a local paths and CMake can use these instead of downloading them if they are provided. It can be added in the future though.

src/json/json_api.h
joehu21: Let's just say "JSON C API".

roshkhatri:
>There will need to be maintainer team for each module that can review each other's PRs. So, at least two maintainers would be useful. Is there anyone besides you @roshkhatri?

Yes, @joehu21  will be the maintainer besides me.

>I'd say let's merge it! Later, we can talk about beta releases and stable releases. The TSC can oversee your release plans and the work on a high level, but I don't see that TSC members have enough time to be deeply involved in these modules.

Yes, once this is merged we can open issues and discuss/work on them.

hwware:
Mostly it looks like fine, agree to merge it. Great work.

hwware (approved):

joehu21 (approved):
LGTM

src/json/json.cc
```diff
@@ -0,0 +1,3232 @@
...
+}
+
+int Config_SetHashTableMinSize(const char *name, long long val, void *privdata, ValkeyModuleString **err) {
+    VALKEYMODULE_NOT_USED(name);
+    VALKEYMODULE_NOT_USED(privdata);
+    VALKEYMODULE_NOT_USED(err);
+    return handleHashTableFactor(&rapidjson::HashTableFactors::minHTSize, &val, size_t(1));
+}
+
+int registerModuleConfigs(ValkeyModuleCtx *ctx) {
```
madolson, line 2540: The RFC has 2 configs, why does this have like 50?

src/json/stats.h
```diff
@@ -0,0 +1,138 @@
+/**
+ * The STATS module's main responsibility is to produce the following metrics:
+ * 1. Core metrics:
+ *    json_total_memory_bytes: total memory allocated to JSON objects
```
madolson, line 4: The original rfc has 2 metrics, why does this one have so much more?

madolson (commented):
Just evaluated the RFC against the implementation. There seems to be some discrepancies, so would like to start conversations about why.

src/json/json.cc
```diff
@@ -0,0 +1,3232 @@
...
+        return false;
+    }
+    return true;
+}
+
+/*
+ * Check the encoding version, For unsupported versions we ALWAYS put out a message in the log
+ * but we only fail the RDB load if the config tells us to do it.
+ */
+bool checkVersion(const char *type_name, int encver, int expected_encver) {
```
madolson, line 2652: I suppose I don't get what this is for. What does this do for compatibility?

madolson, line 2652: I suppose I don't get what this is for. What does this do for compatibility?

src/json/json.cc
```diff
@@ -0,0 +1,3232 @@
...
+    if (strncmp(value, str.get(), str_len)) {
+        ValkeyModule_Log(nullptr, "warning", "%s: Unexpected value in RDB. Expected %s Received %s",
+                        caller, value, str.get());
+        return false;
+    }
+    return true;
+}
+
+/*
+ * Check an integer value, fail
```
madolson: Overall, this code is not very well documented. This comment just trails off for example.

src/json/json.cc
```diff
@@ -0,0 +1,3232 @@
...
+                        DOCUMENT_TYPE_NAME, DOCUMENT_TYPE_ENCODING_VERSION);
+        return VALKEYMODULE_ERR;
+    }
+
+    /*
+     * Now create the stub datatypes for search
+     */
+    if (!install_stub(ctx, "scdtype00", SCDTYPE_ENCVER, scdtype_aux_load)) return VALKEYMODULE_ERR;
+    if (!install_stub(ctx, "GEARS_DT0", GEARSDT_ENCVER, gearsdt_aux_load)) return VALKEYMODULE_ERR;
+    if (!install_stub(ctx, "GEAR_REQ0", GEARSRQ_ENCVER, gearsrq_aux_load)) return VALKEYMODULE_ERR;
```
madolson, line 2830: Apparently we also load 3 custom module types here. I understand this to be for compatibility with stack, but do we want to have this in this module?

src/json/json.cc
```diff
@@ -0,0 +1,3232 @@
...
+    if (ValkeyModule_SetCommandACLCategories(ValkeyModule_GetCommand(ctx,"JSON.DEL"), cat_fast_write) == VALKEYMODULE_ERR) {
+        ValkeyModule_Log(ctx, "warning", "Failed to set command category for JSON.DEL.");
+        return VALKEYMODULE_ERR;
+    }
+
+    if (ValkeyModule_CreateCommand(ctx, "JSON.FORGET", Command_JsonDel, cmdflg_fast_write, 1, 1, 1) == VALKEYMODULE_ERR) {
+        ValkeyModule_Log(ctx, "warning", "Failed to create command JSON.FORGET.");
+        return VALKEYMODULE_ERR;
+    }
+    if (ValkeyModule_SetCommandACLCategories(ValkeyModule_GetCommand(ctx,"JSON.FORGET"), cat_fast_write) == VALKEYMODULE_ERR) {
```
madolson, line 2906: This code doesn't look linted, maybe create a followup issue to make sure we actually running the linter automatically. I guess generally we're missing workflow actions.

madolson (commented):
I am mostly trusting the code, can we setup github workflows to make sure this code all builds and what not?

src/json/json.cc
```diff
@@ -0,0 +1,3232 @@
...
+ * This file implements the Valkey Module interfaces.
+ *
+ * When the module is loaded, it does the following:
+ * 1. register the JSON module.
+ * 2. register callback methods such as rdb_load, rdb_save, free, etc.
+ * 3. register JSON data type.
+ * 4. register commands that are all prefixed with "JSON.".
+ *
+ * Design Considerations:
+ * 1. Command API: see API.md.
```
madolson: 👻 Mystery file in API.md.

madolson:
Ideally let's make issues for the comments here. The only thing I think we should fix before merging is getting the CI working.

src/json/json.cc
zuiderkwast: 5️⃣0️⃣⁉️

It's easy to add configs but it's a breaking change to remove one. Without looking at what they are for, can we replace them with constants or something before we release a version where we commit to maintaining them?

src/json/json.cc
zuiderkwast: Interesting. At least, this should be added to the rfc.

roshkhatri:
I have added the CI workflow to run against valkey `unstable` and `8.0.0` for now.

Screenshot 2024-11-26 at 12 57 52 PM

src/json/json.cc
joehu21: Yes, that's for RDB compatibility with RedisJSON. If we don't care about them any more, we can remove the code.

src/json/json.cc
joehu21: The function is invoked when loading a RDB produced by RedisJSON. Their RDB could include a newer version of a module type. This function decides whether or not to reject the RDB.

src/json/json.cc
joehu21: We considered dropping these configs. However, it requires a decent amount of code cleanup. Agreed we should remove them. We can create an issue to replace these configs with constants.

src/json/json.cc
madolson: Won't it just get rejected if it's from a newer version?

src/json/stats.h
joehu21: These module infos have been removed (not emitted). However, the documentation is forgotten to cleaned up.

src/json/json.cc
joehu21: Yeah, the comment should be "Check an integer value, fail if the value doesn't match the expected one."

src/json/json.cc
joehu21: The line should be removed.

src/json/json.cc
zuiderkwast: I think we do care about compatibility with RedisJSON. We want to make it easy to migrate to Valkey.

I just wish it is mentioned in the RFC, which already has a lot of details.

madolson: These aren't for json compatibility though, they list gears and search types.

zuiderkwast: Which are partially used by the JSON module or vice versa? Or the JSON module provides a minimal subset of them or something? I don't know this. :laughing:

I guess the only reason to add them is to support RDB streams containing module types from these modules. Otherwise, delete it.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,25 @@
...
+
+jobs:
+  build-ubuntu-latest:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        server_version: ['unstable', '8.0.0']
+    steps:
+    - uses: actions/checkout@v4
```
PingXie, line 13: consider using the commit hash for the action instead.

CMakeLists.txt
PingXie: looks like this is all about reusing the tests at https://github.com/valkey-io/valkey-bloom/tree/unstable/tests? or there is more?

CMakeLists.txt
```diff
@@ -0,0 +1,167 @@
...
+    COMMAND tr -d '\n'
+    OUTPUT_VARIABLE ARCHITECTURE
+)
+
+if("${ARCHITECTURE}" STREQUAL "x86_64")
+    message("Building JSON for x86_64")
+elseif("${ARCHITECTURE}" STREQUAL "aarch64")
+    message("Building JSON for aarch64")
+else()
+    message(FATAL_ERROR "Unsupported architecture. JSON is only supported on x86_64 and aarch64.")
```
PingXie: print out the architecture?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,25 @@
...
+
+on: [push, pull_request]
+
+jobs:
+  build-ubuntu-latest:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        server_version: ['unstable', '8.0.0']
```
PingXie, line 11: should we go with the tip of the release branch, which is 8.0? Also I think having older engine versions (like 7.2) at some point would help catch regressions. worst case though we could end up with 6 versions based on our support promise.

CMakeLists.txt
```diff
@@ -0,0 +1,167 @@
...
+endif()
+
+# Set C & C++ standard versions
+set(CMAKE_C_STANDARD 11)
+set(CMAKE_C_STANDARD_REQUIRED True)
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED True)
+
+# Always include debug symbols and optimize the code
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -O2 -g")
```
PingXie: not going with -O3?

CMakeLists.txt
```diff
@@ -0,0 +1,167 @@
...
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O2 -g")
+
+# RapidJSON SIMD optimization
+if("${ARCHITECTURE}" STREQUAL "x86_64")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=nehalem")
+elseif("${ARCHITECTURE}" STREQUAL "aarch64")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=armv8-a")
+else()
+    message(FATAL_ERROR "Unsupported architecture. JSON is only supported on x86_64 and aarch64.")
```
PingXie: nit - print out the actual arch too?

CMakeLists.txt
```diff
@@ -0,0 +1,167 @@
...
+set(RAPIDJSON_BUILD_EXAMPLES OFF CACHE BOOL "Build rapidjson examples" FORCE)
+set(RAPIDJSON_BUILD_DOC OFF CACHE BOOL "Build rapidjson documentation" FORCE)
+
+# Make Rapidjson available
+FetchContent_MakeAvailable(rapidjson)
+
+# Add the src subdirectory for building
+add_subdirectory(src)
+
+# Add the src subdirectory for building
```
PingXie: btw, this comment can be skipped IMO
```suggestion
# Add the tst subdirectory for building
```

README.md
```diff
@@ -0,0 +1,89 @@
+# ValkeyJSON
+
+ValkeyJSON is a C++ Valkey-Module that provides native JSON (JavaScript Object Notation) support for Valkey. The implementation complies with RFC7159 and ECMA-404 JSON data interchange standards. Users can natively store, query, and modify JSON data structures using the JSONPath query language. The query expressions support advanced capabilities including wildcard selections, filter expressions, array slices, union operations, and recursive searches.
+
+ValkeyJSON leverages [RapidJSON](https://rapidjson.org/), a high-performance JSON parser and generator for C++, chosen for its small footprint and exceptional performance and memory efficiency. As a header-only library with no external dependencies, RapidJSON provides robust Unicode support while maintaining a compact memory profile of just 16 bytes per JSON value on most 32/64-bit machines.
+
+## Motivation
+While Valkey core lacks native JSON support, there's significant community demand for JSON capabilities. ValkeyJSON provides a comprehensive open-source solution with extensive JSON manipulation features.
```
PingXie: this motivation belongs more to the rfc. Developers who got here would probably care more about the how-tos IMO.

build.sh
```diff
@@ -0,0 +1,63 @@
...
+SCRIPT_DIR=$(pwd)
+echo "Script Directory: $SCRIPT_DIR"
+
+# Ensure SERVER_VERSION environment variable is set
+if [ -z "$SERVER_VERSION" ]; then
+    echo "WARNING: SERVER_VERSION environment variable is not set. Defaulting to unstable."
+    export SERVER_VERSION="unstable"
+fi
+
+if [ "$SERVER_VERSION" != "unstable" ] && [ "$SERVER_VERSION" != "8.0.0" ] ; then
```
PingXie, line 17: this version check seems too restrictive. this tool is developer facing so might just open it up to all versions/commits

src/json/json.cc
PingXie: yeah these aux data types seem to just play a supporting role so far. can we register a generic stub instead and just log a message that "xxx is skipped" and move on? there could be more aux types added in the future in rejson so it will be good to have that visibility if community members report any issue.

```c
install_stub(ctx, "*", GENERIC_ENCVER, generic_aux_load);
```

src/json/json.cc
```diff
@@ -0,0 +1,3232 @@
...
+    if (!checkVersion("gearsdt", encver, GEARSDT_ENCVER)) return VALKEYMODULE_ERR;
+    if (when == VALKEYMODULE_AUX_AFTER_RDB) {
+        if (!checkString(ctx, "StreamReader", "gears-dt")) return VALKEYMODULE_ERR;
+        if (!checkInt(ctx, 0, "gears-dt")) return VALKEYMODULE_ERR;
+        if (!checkString(ctx, "CommandReader", "gears-dt")) return VALKEYMODULE_ERR;
+        if (!checkInt(ctx, 0, "gears-dt")) return VALKEYMODULE_ERR;
+        if (!checkString(ctx, "KeysReader", "gears-dt")) return VALKEYMODULE_ERR;
+        if (!checkInt(ctx, 0, "gears-dt")) return VALKEYMODULE_ERR;
+        if (!checkString(ctx, "", "gears-dt")) return VALKEYMODULE_ERR;
+    }
```
PingXie, line 2706: these checks add fragility only to the module compat. Since we are not planning to consume any of them, the best strategy imo would be to log a "xxx is skipped" notice level message and return _OK. same goes to other stubs.

PingXie (approved):
This does look overwhelming and I only reviewed haphazardly. I agree we should just merge this PR and polish it incrementally, including adding more test coverage.

Thanks @roshkhatri and @joehu21!

src/json/json.cc
joehu21: Default value of enforce_rdb_version_check is false. So, the check will pass. In our internal RDB compatibility tests, which isn't part of the PR, we set enforce_rdb_version_check to true so that we will be alerted when ReJSON introduced some new aux types. In retrospect, I think the aux type version checking is an overkill. We should just read and ignore these aux types.

src/json/json.cc
joehu21: > These aren't for json compatibility though, they list gears and search types.

When testing compatibility with ReJSON, we use docker image "docker run -p 6389:6379 redis/redis-stack:latest", which loads a bunch of modules including search, graph, etc. Although our test only creates json keys, the generated RDB contains a bunch of aux types like scdtype00, gear*, etc. I suspect these aux types are from RedisSearch module, but not 100% sure because I never looked at their source code. (NOTE: It also indicates that their modules do not use aux_save2, which would not save aux types that have no corresponding data.)

To be able to load the RDB, we register these aux types, read the data and ignored them. So, these are indeed for JSON compatibility test.

src/json/json.cc
joehu21: > yeah these aux data types seem to just play a supporting role so far. can we register a generic stub instead and just log a message that "xxx is skipped" and move on? there could be more aux types added in the future in rejson so it will be good to have that visibility if community members report any issue.

The challenge is that each aux type may have different number of entries in RDB. For example, scdtype only has one integer, while gearsdt has 7 entries in RDB (some are int and some are string).

CMakeLists.txt
joehu21: We want the max optimization that still gives us debug symbols with line numbers in the crash stack. I believe such opt level is -O2.

CMakeLists.txt
joehu21: Yes, printing out `${ARCHITECTURE}` will be helpful.

joehu21: > looks like this is all about reusing the tests at https://github.com/valkey-io/valkey-bloom/tree/unstable/tests? or there is more?

Yes, it's for reusing the pytest infrastructure in valkey-bloom.

CMakeLists.txt
joehu21: Yes, it's all about reusing the pytest infrastructure in valkey-bloom.

src/json/json.cc
PingXie: You are right @joehu21. These aux data types are indeed needed. The main purpose of these handlers is actually to skip these aux types.

CMakeLists.txt
PingXie: hmm this dependency is a bit odd - does it make sense to pull the pytest infra into the Valkey repo instead? We are discussing migrating to python based integration test infra for Valkey too.

CMakeLists.txt
PingXie: maybe at least make is configurable on the command line so it could be overwritten if so desired?

src/json/json.cc
PingXie: Got it - the parsing part makes sense to me now. But maybe we could skip the check of the actual value?

src/json/json.cc
zuiderkwast: If we ever want another module that provides compatibilty with redis search or gears modules, then these names are already registered by ValkeyJSON module so the other modules can't be loaded? That's a potential future problem.

CMakeLists.txt
roshkhatri: Yes, I would suggest we should pull it out in the valkey repo instead. However, we can change the line later when we have it separately.

src/json/json.cc
joehu21: Totally agreed, no need to check the actual value.

CMakeLists.txt
joehu21: agreed. it's desirable to be able to override it through command line argument. @roshkhatri can you create an issue after the PR is merged?

src/json/json.cc
joehu21: That's a good point.

madolson:
Before we merge, please just make sure everything has issues so we can track them.

madolson (approved):

enjoy-binbin (approved):
did not review, but it is indeed a good work, thanks.

CMakeLists.txt
madolson: There are still debug symbols with a crash at -O3, you might be thinking about frame pointers which allow unwinding the stack. You can keep those with `-fno-omit-frame-pointer`.

> @roshkhatri can you create an issue after the PR is merged?

Please create the issues before the PR is merged. It's better practice to define the follow up work before merging in the current code.

CMakeLists.txt
madolson: I was going to suggest we write an RFC or something similar for the python testing framework as well before we get too far along building it organically.

src/json/json.cc
madolson: Yeah, I think we need a better overall strategy here instead of install random types and ignoring the data from them.

src/json/json.cc
roshkhatri: https://github.com/valkey-io/valkeyJSON/issues/2

src/json/json.cc
roshkhatri: https://github.com/valkey-io/valkeyJSON/issues/3

roshkhatri: Added the CI to run all the tests for now.

.github/workflows/ci.yml
roshkhatri: https://github.com/valkey-io/valkeyJSON/issues/4

.github/workflows/ci.yml
roshkhatri: https://github.com/valkey-io/valkeyJSON/issues/5

CMakeLists.txt
roshkhatri: https://github.com/valkey-io/valkeyJSON/issues/6

build.sh
roshkhatri: https://github.com/valkey-io/valkeyJSON/issues/7

src/json/json.cc
roshkhatri: https://github.com/valkey-io/valkeyJSON/issues/8

src/json/json.cc
roshkhatri: https://github.com/valkey-io/valkeyJSON/issues/9

roshkhatri:
I have created relevant issues for the comments and resolved those conversations to track them better. I have also addressed some feedback for minor changes.

Let me know if any more issues are required to be opens before we merge this PR. Happy Thanksgiving Everyone!!

soloestoy (approved):

CMakeLists.txt
zuiderkwast: Why not move the testing framework to it's own repo to start with? It only needs to be pulled in for running the tests, not when just building the module.

I'd rather see people spend time on a README for the test framework than on an RFC. Or we can do both in parallel...?

hwware:
> I have created relevant issues for the comments and resolved those conversations to track them better. I have also addressed some feedback for minor changes.
> 
> Let me know if any more issues are required to be opens before we merge this PR. Happy Thanksgiving Everyone!!

Happy Black Friday !

CMakeLists.txt
madolson: > I'd rather see people spend time on a README for the test framework than on an RFC. Or we can do both in parallel...?

I just want us to have some long term plan for testing and not just build it up organically. The reason is that Amazon internally has a similar testing framework that is extremely complex because nobody really maintains it and just bolt on random features. There are many cases where it's possible to do the same way 2 or 3 different ways. I think the same is true of our TCL framework, just to a smaller scale. 

So whether it be an RFC or just some direction in a new repo, I don't care which.

CMakeLists.txt
madolson: https://github.com/valkey-io/valkeyJSON/issues/10

src/json/stats.h
madolson: https://github.com/valkey-io/valkeyJSON/issues/11

src/json/json.cc
madolson: https://github.com/valkey-io/valkeyJSON/issues/12

madolson:
We have 6/6 TSC actually approved, so going ahead and merging it.

PR #13, opened 2024-11-29
Update integration test script

joehu21:
Enhance the integration test script to check if environment variable SERVER_VERSION is set. If not, default to "unstable". Also update the README for how to run one integration test.

Details:

1. Running the integration test using "make test" failed:
```
Running integration tests against Valkey version: 
.build/binaries//valkey-server missing
make[3]: *** [CMakeFiles/test] Error 1
make[2]: *** [CMakeFiles/test.dir/all] Error 2
make[1]: *** [CMakeFiles/test.dir/rule] Error 2
make: *** [test] Error 2
```
2. The root cause is that environment variable SERVER_VERSION is missing.
3. Fixed by checking if environment variable SERVER_VERSION is set. If not, default to "unstable".

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,16 @@ make -j unit
 make -j test
 ```
 
+#### To run one integration test:
+```text
+TEST_PATTERN=<test-function-or-file> make -j test
+```
+e.g.,
+```text
+TEST_PATTERN=test_sanity make -j test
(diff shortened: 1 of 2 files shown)


roshkhatri (approved):
Looks good to me!

PR #15, opened 2024-12-04
Fix #12 and #8 - Simplify check for aux types in RDB version

### LINKED ISSUE #8: Skip the check of the actual value of aux types
```
Got it - the parsing part makes sense to me now. But maybe we could skip the check of the actual value?
```
_Originally posted by @PingXie in https://github.com/valkey-io/valkeyJSON/pull/1#discussion_r1859995125_

### LINKED ISSUE #12: Simplify check for aux types in RDB version
Default value of enforce_rdb_version_check is false. So, the check will pass. In our internal RDB compatibility tests, which isn't part of the PR, we set enforce_rdb_version_check to true so that we will be alerted when ReJSON introduced some new aux types. In retrospect, I think the aux type version checking is an overkill. We should just read and ignore these aux types.

_Originally posted by @joehu21 in https://github.com/valkey-io/valkeyJSON/pull/1#discussion_r1859771586_

joehu21:
Fix issue #12 and #8. There is no need to check redis stack's aux types versions. Neither is there a need to check values of the aux type data. We just need to load the data and ignore.

### DIFF
--- a/src/json/json.cc
+++ b/src/json/json.cc
@@ -73,8 +73,6 @@ KeyTable *keyTable = nullptr;
 rapidjson::HashTableFactors rapidjson::hashTableFactors;
 rapidjson::HashTableStats   rapidjson::hashTableStats;
 
-bool enforce_rdb_version_check = false;
-
 extern size_t hash_function(const char *text, size_t length);
 
 size_t json_get_max_document_size() {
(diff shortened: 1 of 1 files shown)


roshkhatri (approved):
LGTM!

Issue #20, opened 2024-12-06
ValkeyJSON is not RDB compatible with ReJSON 2.8 or later


joehu21:
Up to ReJSON v2.6.12, ValkeyJSON is RDB compatible with ReJSON, after issue #19 is resolved. Starting ReJSON v2.8, ReJSON had an RDB format change that broke the compatibility.

Trying to load a ReJSON v2.8.3 RDB using ValkeyJSON:
```
  7131:M 06 Dec 2024 17:12:06.150 * Module 'json' loaded from /home/jowhuw/github/joehu21-valkeyJSON/build/src/libjson.so
  7131:M 06 Dec 2024 17:12:06.150 * Server initialized
  7131:M 06 Dec 2024 17:12:06.150 # Can't handle RDB format version 12
  7131:M 06 Dec 2024 17:12:06.150 # Fatal error loading the DB, check server logs. Exiting.
```

Recommendation:
For now, we should at least update our documentation to state that ValkeyJSON is RDB compatible with ReJSON up to v2.6.12. It's unknown if we can still maintain the compatibility with ReJSON v2.8 and later.

roshkhatri:
Backward compatibility till 2.6.12 is resolved by https://github.com/valkey-io/valkeyJSON/issues/19

We need to discuss if we want to be compatible with ReJSON v2.8

joehu21:
This is actually expected. ReJSON v2.8 uses redis 7.2.1. ValkeyJSON makes no commit of RDB compatibility with ReJSON beyond Redis 7.2, which is no longer an open source.

joehu21:
Closed the issue because ValkeyJSON makes no commit of RDB compatibility with ReJSON v2.8 or later that is beyond the last OSS Redis version - 7.1. ValkeyJSON is only RDB-compatible with ReJSON v1.0.8 to v2.6.12.

joehu21:
Retested. We are actually RDB compatible with RedisJSON v1.0.8 or later.

PR #21, opened 2024-12-06
Fix #19 - RDB compatibility with ReJSON and compatibility test

### LINKED ISSUE #19: RDB compatibility with ReJSON
Found a bug that RDB compatibility with ReJSON is broken. The root cause is that some ReJSON aux type processing code were missed out in the initial PR. The bug was not caught in the integration test because the RDB compatibility test is missing. So, we should fix the test gap as well.

Also discovered that we are only RDB compatible with ReJSON up to ReJSON v2.6.12. ReJSON has changed RDB format since v2.8. Created a separate issue: #20.

joehu21:
Fixed RDB compatibility with ReJSON (up to v2.6.12), and added RDB compatibility test.


tst/integration/rdb_rejson/README.txt
roshkhatri, line 1: Why do we not make this README.md to keep it uniforms across the repository

tst/integration/rdb_rejson/README.txt
```diff
@@ -0,0 +1,14 @@
+How to create rdb file for a new ReJSON release?
+
+e.g., testing RDB compatibility with rejson 2.2.0.
+
+1. Run docker image redis-stack:
+    docker run -d -p 6379:6379 --name rejson redislabs/rejson:2.2.0
```
roshkhatri:
```suggestion
    ```text
    docker run -d -p 6379:6379 --name rejson redislabs/rejson:2.2.0
    ```
```

tst/integration/rdb_rejson/README.txt
```diff
@@ -0,0 +1,14 @@
...
+2. Load store.json and create a key named "store":
+    python3 utils/load_1file_hostport.py tst/integration/data/store.json store
```
roshkhatri:
```suggestion
    ```text
    python3 utils/load_1file_hostport.py tst/integration/data/store.json store
    ```
```

tst/integration/rdb_rejson/README.txt
```diff
@@ -0,0 +1,14 @@
...
+3. Save rdb:
+    valkey-cli save
```
roshkhatri:
```suggestion
    ```text
    valkey-cli save
    ```
```

tst/integration/rdb_rejson/README.txt
```diff
@@ -0,0 +1,14 @@
...
+4. Copy out the RDB file:
+    docker cp $(docker ps -q):/data/dump.rdb tst/integration/rdb_rejson/rejson-<version>.rdb
```
roshkhatri: similarly in code blocks

tst/integration/rdb_rejson/README.txt
```diff
@@ -0,0 +1,14 @@
...
+5. run test_json_rdb_import.py:
+    TEST_PATTERN=test_import_rejson_rdbs make test
```
roshkhatri: similarly in code blocks

roshkhatri (commented):
Everything else looks great, just some changes

tst/integration/rdb_rejson/README.txt
joehu21: changed to README.md

tst/integration/rdb_rejson/README.txt
joehu21: updated

tst/integration/rdb_rejson/README.txt
joehu21: updated

tst/integration/rdb_rejson/README.txt
joehu21: updated

tst/integration/rdb_rejson/README.txt
joehu21: updated

roshkhatri (approved):
LGTM

src/json/json.cc
```diff
@@ -2527,6 +2547,8 @@ extern "C" int ValkeyModule_OnLoad(ValkeyModuleCtx *ctx) {
      * Now create the stub datatypes for search
      */
     if (!install_stub(ctx, "scdtype00", SCDTYPE_ENCVER, scdtype_aux_load)) return VALKEYMODULE_ERR;
+    if (!install_stub(ctx, "ft_index0", FTINDEX_ENCVER, ftindex_aux_load)) return VALKEYMODULE_ERR;
```
madolson, line 2550: I asked to remove these, not add more. I'm trying to take a step back and understand if we need to include a better way to ignore this auxiliary data. This is the JSON module, this is not the search, gears, whatever module. This feels like a hack.

joehu21: The following is a test to explain why we need to load search aux types:
1. Start ReJSON docker image:
```
docker run -p 6379:6379 --name redis-redisjson redislabs/rejson:2.0.6
```
Although it's a rejson image, it contains search module as well:
```
127.0.0.1:6379> module list
1) 1) "name"
   2) "ReJSON"
   3) "ver"
   4) (integer) 20006
2) 1) "name"
   2) "search"
   3) "ver"
   4) (integer) 20205
```
2. We cannot unload the search module because it exports the search aux types:
```
127.0.0.1:6379> module unload search
(error) ERR Error unloading module: the module exports one or more module-side data types, can't unload
```
4. create one json key:
```
json.set k1 . 1
```
5. save
6. copy out RDB:
```
docker cp $(docker ps -q):/data/dump.rdb dump.rdb
```
5. load the RDB using ValkeyJSON binary that does not include PR #21. It failed:
```
22242:M 12 Dec 2024 02:50:45.568 * Loading RDB produced by Redis version 6.2.5
22242:M 12 Dec 2024 02:50:45.568 * RDB age 124 seconds
22242:M 12 Dec 2024 02:50:45.568 * RDB memory usage when created 0.92 Mb
22242:M 12 Dec 2024 02:50:45.568 # The RDB file contains AUX module data I can't load: no matching module 'ft_index0'
```
That's because code for loading aux type ft_index0 is missing.
6. load the RDB using ValkeyJSON binary that includes PR #21. It succeeded.

Note that although we have not created any search index, Redis still writes search aux types to RDB. I think the problem comes from RediSearch. It must have used aux_save instead of aux_save2. For aux_save, even if there is no module data (search index in this case), it writes aux type meta data to RDB. The correct way is to use aux_save2, which will not write module types if there is no module data. The fix should come from RediSearch. We can open an issue for them. If they fix the problem, then we do not need to load search aux types.


I just realized some of the rejson RDB test files, included in this PR, were generated using `redis-stack` docker images and that's why gears and graph aux types need to be registered in order to load them. This can be fixed by regenerating the RDBs using rejson docker images, and remove code related to loading grears & graph aux types.

However, there's still a question regarding search aux types. If we want to help migrating ReJSON to VakeyJSON, do we need to load aux search types?

I believe there are two search aux types - scdtype00 and ft_index00. Some ReJSON docker images contain RediSearch and some do not. e.g.,
1. The following images contain RediSearch:
  * docker run -p 6380:6379 --name redis-redisjson redislabs/rejson:latest
  * docker run -p 6379:6379 --name redis-redisjson redislabs/rejson:2.0.6
  * any release of v2.0 or older
2. The followings do not contain RediSearch (between v 2.2 and v2.6.12):
  * docker run -p 6379:6379 --name redis-redisjson redislabs/rejson:2.6.12
  * docker run -p 6379:6379 --name redis-redisjson redislabs/rejson:2.4.0
  * docker run -p 6379:6379 --name redis-redisjson redislabs/rejson:2.2.0  
(NOTE: we are not compatible with ReJSON v2.8 or later anyway. See #20. So it's not listed here.)

For RDBs generated from docker images that do not contain RediSearch, we do not need to load search aux types. However, for RDBs generated from docker images that contain RediSearch, we need to load the search aux types.

PR #39, opened 2025-02-27
Adds JSON.MSET command

### LINKED ISSUE #16: Implement JSON.MSET
Implement JSON.MSET, as specified in the RFC: https://github.com/valkey-io/valkey-rfc/blob/main/ValkeyJSON.md#jsonmset.

#### JSON.MSET

Set JSON values for multiple keys. The operation is atomic. Either all values are set or none is set.

##### Syntax

```bash
JSON.MSET <key> <path> <value> [key <path> <value> ...]
```

* key - required, JSON key
* path - required, a JSON path
* value - JSON value

##### Return

* Simple String 'OK' on success
* Error on failure

roshkhatri:
fixes https://github.com/valkey-io/valkey-json/issues/16

Implements JSON.MSET, as specified in the RFC: https://github.com/valkey-io/valkey-rfc/blob/main/ValkeyJSON.md#jsonmset.

#### JSON.MSET
Set JSON values for multiple keys. The operation is atomic. Either all values are set or none is set.

#### Syntax
```
JSON.MSET <key> <path> <value> [key <path> <value> ...]
```

* key - required, JSON key
* path - required, a JSON path
* value - required, a JSON value
#### Return
* Simple String 'OK' on success
* Error on failure


src/json/json.cc
```diff
@@ -262,6 +262,107 @@ STATIC JsonUtilCode parseSetCmdArgs(ValkeyModuleString **argv, const int argc, S
...
+    // Validate that the number of arguments is correct for MSET (3 arguments per key)
+    if ((argc - 1) % 3 != 0) {
+        return JSONUTIL_WRONG_NUM_ARGS;
+    }
+
+    *num_keys = (argc - 1) / 3;
+    // Allocate memory for args_list 
+    *args_list = reinterpret_cast<MSetCmdArgs *>(ValkeyModule_Alloc((*num_keys) * sizeof(MSetCmdArgs)));
+    if (!(*args_list)) {
+        return JSONUTIL_WRONG_NUM_ARGS; //JSONUTIL_ALLOCATION_FAILURE
```
joehu21: wrong error code

joehu21: Need to zero out *args_list.

src/json/json.cc
```diff
@@ -262,6 +262,107 @@ STATIC JsonUtilCode parseSetCmdArgs(ValkeyModuleString **argv, const int argc, S
...
+    for (i = 0; i < *num_keys; ++i) {
+        MSetCmdArgs &current_arg = (*args_list)[i];
+
+        current_arg.key_str = argv[i * 3 + 1];
+        current_arg.key = static_cast<ValkeyModuleKey*>(
+            ValkeyModule_OpenKey(ctx, current_arg.key_str, VALKEYMODULE_READ | VALKEYMODULE_WRITE));
+
+        // Handle key allocation failure
+        if (!current_arg.key) {
+            rc = JSONUTIL_WRONG_NUM_ARGS; // JSONUTIL_KEY_OPEN_ERROR
```
joehu21: wrong error code

src/json/json.cc
```diff
@@ -262,6 +262,107 @@ STATIC JsonUtilCode parseSetCmdArgs(ValkeyModuleString **argv, const int argc, S
...
+            goto cleanup;
```
joehu21: "goto" is a bad programming style. Remove goto. No need to do cleanup here. Since we return error code to the caller. The caller should check error code. If it is an error, the caller frees the arg_list.

src/json/json.cc
```diff
@@ -262,6 +262,107 @@ STATIC JsonUtilCode parseSetCmdArgs(ValkeyModuleString **argv, const int argc, S
...
+
+        current_arg.path = ValkeyModule_StringPtrLen(argv[i * 3 + 2], nullptr);
+        current_arg.json = ValkeyModule_StringPtrLen(argv[i * 3 + 3], &current_arg.json_len);
+
+        // Validate key type
+        int type = ValkeyModule_KeyType(current_arg.key);
+        if (type != VALKEYMODULE_KEYTYPE_EMPTY &&
+            ValkeyModule_ModuleTypeGetType(current_arg.key) != DocumentType) {
+            rc = JSONUTIL_NOT_A_DOCUMENT_KEY;
+            goto cleanup;
```
joehu21: same here and all other "goto"s.

src/json/json.cc
```diff
@@ -262,6 +262,107 @@ STATIC JsonUtilCode parseSetCmdArgs(ValkeyModuleString **argv, const int argc, S
...
+            }
+            rc = dom_verify_value(ctx, doc, current_arg.path, current_arg.json);
+            if (rc != JSONUTIL_SUCCESS) {
+                goto cleanup;
+            }
+        }
+    }
+    return JSONUTIL_SUCCESS;
+
+cleanup:
```
joehu21: Remove the cleanup section. Let the caller do the cleanup.

src/json/json.cc
```diff
@@ -262,6 +262,107 @@ STATIC JsonUtilCode parseSetCmdArgs(ValkeyModuleString **argv, const int argc, S
...
+    // Free allocated memory and close opened keys
+    for (size_t j = 0; j <= i; ++j) {
+        if ((*args_list)[j].key) {
+            ValkeyModule_CloseKey((*args_list)[j].key);
```
joehu21: This is problematic. Remove L356-360.

src/json/json.cc
```diff
@@ -262,6 +262,107 @@ STATIC JsonUtilCode parseSetCmdArgs(ValkeyModuleString **argv, const int argc, S
...
+        if (is_new_valkey_key && !current_arg.is_root_path) {
+            rc = JSONUTIL_COMMAND_SYNTAX_ERROR;
+            goto cleanup;
+        }
+
+        // Validate JSON structure
+        if (current_arg.is_root_path) {
+            JDocument *doc = nullptr;
+            rc = dom_parse(ctx, current_arg.json, current_arg.json_len, &doc);
+            if (rc != JSONUTIL_SUCCESS) {
```
joehu21, line 326: remove these 3 lines. Just assert rc is success.

src/json/json.cc
roshkhatri: The whole cleanup section is removed

src/json/json.cc
```diff
@@ -629,6 +730,76 @@ int Command_JsonSet(ValkeyModuleCtx *ctx, ValkeyModuleString **argv, int argc) {
...
 }
 
+int Command_JsonMSet(ValkeyModuleCtx *ctx, ValkeyModuleString **argv, int argc) {
+    ValkeyModule_AutoMemory(ctx);
+
+    MSetCmdArgs *args_list = nullptr;
+    size_t num_keys;
+    JsonUtilCode rc = parseAndValidateMSetCmdArgs(ctx, argv, argc, &args_list, &num_keys);
+    if (rc != JSONUTIL_SUCCESS) {
+        return ValkeyModule_ReplyWithError(ctx, jsonutil_code_to_message(rc));
```
joehu21: Here, free args_list.

joehu21: change L740 to:
```
if (rc == JSONUTIL_WRONG_NUM_ARGS)
      return ValkeyModule_WrongArity(ctx);
else
      return ValkeyModule_ReplyWithError(ctx, jsonutil_code_to_message(rc));
```

src/json/json.cc
```diff
@@ -629,6 +730,76 @@ int Command_JsonSet(ValkeyModuleCtx *ctx, ValkeyModuleString **argv, int argc) {
...
+        ValkeyModule_NotifyKeyspaceEvent(ctx, VALKEYMODULE_NOTIFY_GENERIC, "json.mset", args_list[i].key_str);
+        ValkeyModule_CloseKey(args_list[i].key);
+    }
+
+    // replicate the entire command
+    ValkeyModule_ReplicateVerbatim(ctx);
+    ValkeyModule_Free(args_list);
+    return ValkeyModule_ReplyWithSimpleString(ctx, "OK");
+
+cleanup:
```
joehu21: same here, do not use "cleanup" style.

src/json/json.cc
```diff
@@ -629,6 +730,76 @@ int Command_JsonSet(ValkeyModuleCtx *ctx, ValkeyModuleString **argv, int argc) {
...
+            int64_t delta = jsonstats_end_track_mem(begin_val);
+            size_t new_doc_size = dom_get_doc_size(doc) + delta;
+            dom_set_doc_size(doc, new_doc_size);
+
+            // update stats
+            jsonstats_update_stats_on_update(doc, orig_doc_size, new_doc_size, args_list[i].json_len);
+        }
+
+        ValkeyModule_NotifyKeyspaceEvent(ctx, VALKEYMODULE_NOTIFY_GENERIC, "json.mset", args_list[i].key_str);
+        ValkeyModule_CloseKey(args_list[i].key);
```
joehu21: No need to close key because we are using auto memory management.

src/json/json.cc
```diff
@@ -262,6 +262,95 @@ STATIC JsonUtilCode parseSetCmdArgs(ValkeyModuleString **argv, const int argc, S
...
+        if (is_new_valkey_key && !current_arg.is_root_path) {
+            rc = JSONUTIL_COMMAND_SYNTAX_ERROR;
+            return rc;
+        }
+
+        // Validate JSON structure
+        if (current_arg.is_root_path) {
+            JDocument *doc = nullptr;
+            rc = dom_parse(ctx, current_arg.json, current_arg.json_len, &doc);
+            ValkeyModule_Assert(rc == JSONUTIL_SUCCESS);
```
joehu21: Cannot use assert here because the json input string could be invalid. Need to return the error code. Change the line to:
```
if (rc != JSONUTIL_SUCCESS) {
  if (doc) dom_free_doc(doc);
   return rc;
}
```

hpatro (commented):
Could we add test case to verify cross slot operations are disallowed?

joehu21 (approved):

roshkhatri:
> Could we add test case to verify cross slot operations are disallowed?

The current pytest framework does not support cluster mode, it starts a single node in standalone mode. We can add the tests later once we start supporting cluster mode enabled. It isn't a trivial task to add that functionality to pytest, would require a PR of its own.

roshkhatri:
Just checked manually:
```
127.0.0.1:6379> json.mset k1 .age 30 k1 .address.city '"Boston"' k1{hahaha} .phoneNumbers[0].type '"work"'
(error) CROSSSLOT Keys in request don't hash to the same slot
127.0.0.1:6379> json.mset k1 . '{"firstName":"John","lastName":"Smith","age":27,"weight":135.25,"isAlive":true,"address":{"street":"21 2nd Street","city":"New York","state":"NY","zipcode":"10021-3100"},"phoneNumbers":[{"type":"home","number":"212 555-1234"},{"type":"office","number":"646 555-4567"}],"children":[],"spouse":null}'
OK
127.0.0.1:6379> json.get k1 .age
"27"
127.0.0.1:6379> JSON.MSET k1 . '"Boston"' k2 . '123' k3 . '["Seattle","Boston"]' k3 . '[1,2,3]' k4 . '{"a":"b"}' k4 . '{}' k5 . 'null' k6 . 'false'
(error) CROSSSLOT Keys in request don't hash to the same slot
127.0.0.1:6379> JSON.MSET k1 . '"Boston"' k2{k1} . '123'
OK
127.0.0.1:6379> json.get k1 
"\"Boston\""
127.0.0.1:6379> json.get k2{k1}
"123"
```
CROSSSLOT opperations are disallowed from the server.

PR #41, opened 2025-03-05
Resolved issues when using GCC >= 12

### LINKED ISSUE #17: GCC <=12 is incompatible - CI is biased
Hi,

I tried to compile a container image with the module from `valkey/valkey:8` (not yet on alpine).

A lot of errors happen.

I fixed this by downgrading GCC and G++ to v11.

Here is the Dockerfile that works:

```dockerfile
FROM docker.io/valkey/valkey:8 as build

ARG SERVER_VERSION=8.0.0
RUN set -ex; \
  apt-get update; \
  apt-get install -y build-essential cmake git gcc-11 g++-11 pkg-config; \
  update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-11 1; \
  update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-11 1; \
  mkdir -p /opt;

WORKDIR /opt
RUN set -ex; \
  git clone https://github.com/valkey-io/valkeyJSON.git; \
  mkdir -p /opt/valkeyJSON/build;

WORKDIR /opt/valkeyJSON/build
RUN set -ex; \
  cmake .. -DVALKEY_VERSION=${SERVER_VERSION}; \
  make

FROM docker.io/valkey/valkey:8
RUN mkdir -p /usr/lib/valkey
COPY --from=build /opt/valkeyJSON/build/src/libjson.so /usr/lib/valkey/libjson.so
CMD ["valkey-server", "--loadmodule", "/usr/lib/valkey/libjson.so"]
```

On Github Actions, Ubuntu version is not the latest, and not the one used by the main OCI Image. So the tests are OK on GitHub actions, but not with the official image.

Nikhil-Manglore:
This PR implements a standards-compliant custom allocator for ValkeyJSON that resolves compatibility issues with GCC 13. Additionally I fixed compiler errors and warnings that are related to standards enforced in GCC 13.

Resolves #17

### DIFF
--- a/src/json/dom.cc
+++ b/src/json/dom.cc
@@ -1416,13 +1416,8 @@ STATIC JValue readLegacyDoubleAsJValue(ValkeyModuleIO *rdb) {
     double d = ValkeyModule_LoadDouble(rdb);
     char str[BUF_SIZE_DOUBLE_JSON];
     size_t str_len = jsonutil_double_to_string(d, str, sizeof(str));
-    if (str) {
         JValue v(str, str_len, allocator, false, /* isdouble */ true);
         return v;
-    } else {
-        ValkeyModule_LogIOError(rdb, "error", "Unable to read legacy double");
-        return JValue();
-    }
 }
 
 /*
--- a/src/json/memory.h
+++ b/src/json/memory.h
@@ -98,7 +98,6 @@ namespace jsn
 template <typename T> class stl_allocator {
     public:
         using value_type = T;
-
         stl_allocator() = default;
         stl_allocator(const stl_allocator<T>&) noexcept = default;
         stl_allocator(stl_allocator<T>&&) noexcept = default;
--- a/src/json/memory.h
+++ b/src/json/memory.h
@@ -95,17 +95,40 @@ namespace jsn
 //
 // Our custom allocator
 //
-template <typename T> class stl_allocator : public std::allocator<T> {
- public:
-    typedef T value_type;
-    stl_allocator() = default;
-    stl_allocator(std::allocator<T>&) {}
-    stl_allocator(std::allocator<T>&&) {}
-    template <class U> constexpr stl_allocator(const stl_allocator<U>&) noexcept {}
+template <typename T> class stl_allocator {
+    public:
+        using value_type = T;
 
-    T *allocate(std::size_t n) { return static_cast<T *>(memory_alloc(n*sizeof(T))); }
-    void deallocate(T *p, std::size_t n) { (void)n; memory_free(p); }
+        stl_allocator() = default;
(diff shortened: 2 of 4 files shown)


joehu21 (approved):
LGTM

src/json/dom.cc
```diff
@@ -1416,13 +1416,8 @@ STATIC JValue readLegacyDoubleAsJValue(ValkeyModuleIO *rdb) {
     double d = ValkeyModule_LoadDouble(rdb);
     char str[BUF_SIZE_DOUBLE_JSON];
     size_t str_len = jsonutil_double_to_string(d, str, sizeof(str));
-    if (str) {
```
roshkhatri, line 1419: Why are we removing this check?

I think we should keep this in compliance with GCC 13

src/json/dom.cc
```diff
@@ -1416,13 +1416,8 @@ STATIC JValue readLegacyDoubleAsJValue(ValkeyModuleIO *rdb) {
...
         JValue v(str, str_len, allocator, false, /* isdouble */ true);
         return v;
-    } else {
-        ValkeyModule_LogIOError(rdb, "error", "Unable to read legacy double");
-        return JValue();
-    }
```
roshkhatri, line 1425: Also keep the else block

src/json/memory.h
```diff
@@ -95,18 +95,41 @@ namespace jsn
...
-    stl_allocator(std::allocator<T>&) {}
-    stl_allocator(std::allocator<T>&&) {}
-    template <class U> constexpr stl_allocator(const stl_allocator<U>&) noexcept {}
-
-    T *allocate(std::size_t n) { return static_cast<T *>(memory_alloc(n*sizeof(T))); }
-    void deallocate(T *p, std::size_t n) { (void)n; memory_free(p); }
+template <typename T> class stl_allocator {
+    public:
+        using value_type = T;
+
```
roshkhatri:
```suggestion
```

src/json/dom.cc
Nikhil-Manglore: I think we need to remove it due to stricter checking requirements in GCC13. If I add the original code back in the build fails due to all warnings being treated as errors. The warning arises because the string will never be null.

``` 
error: the address of ‘str’ will never be NULL [-Werror=address]
 1419 |     if (str) {
 ```

src/json/dom.cc
Nikhil-Manglore: Same reason as above.

src/json/dom.cc
joehu21: Yeah, the change LGTM.

roshkhatri (approved):

PR #45, opened 2025-03-13
Enable tests to run with ASAN build.

roshkhatri:
This PR enables ASAN and also also checks for memory leaks in integration tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,3 +23,26 @@ jobs:
         pip install -r requirements.txt
     - name: Build and Run tests. 
       run: ./build.sh
+
+  build-asan-ubuntu-latest:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        server_version: ['unstable', '8.0']
+    steps:
+    - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+    - name: Set the server version for python integration tests
+      run: echo "SERVER_VERSION=${{ matrix.server_version }}" >> $GITHUB_ENV
+    - name: Set up Python
+      uses: actions/setup-python@3542bca2639a428e1796aaa6a2ffef0c0f575566 #v3.1.4
+      with:
+        python-version: '3.9'
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install -r requirements.txt
+    - name: Build and Run tests. 
+      run: |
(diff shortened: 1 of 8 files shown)


tst/integration/run.sh
```diff
@@ -34,7 +34,32 @@ if [[ ! -f "${BINARY_PATH}" ]] ; then
 fi
 
 if [[ $1 == "test" ]] ; then
-    python -m pytest --html=report.html --cache-clear -v ${TEST_FLAG} ./ ${TEST_PATTERN}
+    if [ ! -z "${ASAN_BUILD}" ]; then
+        echo "Running tests and checking for memory leaks"
+        python -m pytest --capture=sys --html=report.html --cache-clear -v ${TEST_FLAG} ./ ${TEST_PATTERN} 2>&1 | tee test_output.tmp
+        # Check for memory leaks in the output
+        if grep -q "LeakSanitizer: detected memory leaks" test_output.tmp; then
```
joehu21, line 41: Detecting memory leak is one of ASAN features. What about other errors detected by ASAN?

joehu21 (commented):
libjson.so is now built with ASAN enabled when ASAN_BUILD is true. What about valkey-server binary? Pytest starts the valkey-server. We also need to use the asan-enabled valkey-server binary when ASAN_BUILD is true.

.github/workflows/ci.yml
```diff
@@ -22,4 +22,27 @@ jobs:
         python -m pip install --upgrade pip
         pip install -r requirements.txt
     - name: Build and Run tests. 
-      run: ./build.sh
\ No newline at end of file
+      run: ./build.sh
+
+  build-asan-ubuntu-latest:
```
joehu21, line 27: When is this target invoked?

roshkhatri: This is invoked on pull requests and on push

roshkhatri:
>libjson.so is now built with ASAN enabled when ASAN_BUILD is true. What about valkey-server binary? Pytest starts the valkey-server. We also need to use the asan-enabled valkey-server binary when ASAN_BUILD is true.

We also build `valkey-server` with ASAN
```
if(ENABLE_ASAN)
    message("Building Valkey engine with Address Sanitizer enabled")
    set(BUILD_CMD make distclean && make -j SANITIZER=address)
else()
    set(BUILD_CMD make distclean && make -j)
endif()
```

joehu21 (commented):
We can mark the failed test as xfail and create a issue to resolve it later.

joehu21 (approved):
LGTM

PR #48, opened 2025-03-20
Add module release stage

roshkhatri:
This is to differentiate between unstable(dev), release candidate(rc) and stable version release(ga) stage

### DIFF
--- a/src/json/json.cc
+++ b/src/json/json.cc
@@ -38,7 +38,14 @@
 #include <memory>
 #include <cmath>
 
-#define MODULE_VERSION 90
+/* In unstable branch the module version is always "000000". */
+#define MODULE_VERSION 000000
+#define MODULE_NAME "json"
+/* The release stage is used in order to provide release status information.
+ * In unstable branch the status is always "dev".
+ * During release process the status will be set to rc1,rc2...rcN.
+ * When the version is released the status will be "ga". */
+#define MODULE_RELEASE_STAGE "dev"
 #define MODULE_NAME "json"
 #define DOCUMENT_TYPE_NAME "ReJSON-RL"
 #define DOCUMENT_TYPE_ENCODING_VERSION 3   /* Currently support 1 or 3 */


joehu21 (approved):
LGTM

src/json/json.cc
```diff
@@ -38,7 +38,14 @@
 #include <memory>
 #include <cmath>
 
-#define MODULE_VERSION 90
+/* In unstable branch the module version is always "000000". */
+#define MODULE_VERSION 000000
```
KarthikSubbarao, line 42: Question: Do we want to use 000000000 or 255255255? 

In Valkey core, we can see the following:
```
31934:M 20 Mar 2025 18:07:48.020 * Valkey version=255.255.255, bits=64, commit=d3c8681b, modified=1, pid=31934, just started
31934:M 20 Mar 2025 18:07:48.020 * Configuration loaded
31934:M 20 Mar 2025 18:07:48.023 * Increased maximum number of open files to 10032 (it was originally set to 2560).
31934:M 20 Mar 2025 18:07:48.023 * monotonic clock: POSIX clock_gettime
                .+^+.                                                
            .+#########+.                                            
        .+########+########+.           Valkey 255.255.255 (d3c8681b/1) 64 bit
    .+########+'     '+########+.                                    
 .########+'     .+.     '+########.    Running in standalone mode
 |####+'     .+#######+.     '+####|    Port: 7000
 |###|   .+###############+.   |###|    PID: 31934                     
 |###|   |#####*'' ''*#####|   |###|                                 
 |###|   |####'  .-.  '####|   |###|                                 
 |###|   |###(  (@@@)  )###|   |###|          https://valkey.io      
 |###|   |####.  '-'  .####|   |###|                                 
 |###|   |#####*.   .*#####|   |###|                                 
 |###|   '+#####|   |#####+'   |###|                                 
 |####+.     +##|   |#+'     .+####|                                 
 '#######+   |##|        .+########'                                 
    '+###|   |##|    .+########+'                                    
        '|   |####+########+'                                        
             +#########+'                                            
                '+v+'          
```

With this change, MODULE LIST and the HELLO command will show the `ver` as 0.

joehu21: The reason that valkey-json version uses 6-digit is as follows: The intention is to be compatible with Redis Lab's module version scheme. All of RL modules use 6-digit version number. e.g., RedisJSON module v2.8.8:
```
3) 1) "name"
   2) "ReJSON"
   3) "ver"
   4) (integer) 20808
   5) "path"
   6) "/opt/redis-stack/lib/rejson.so"
   7) "args"
   8) (empty array)
```
I'm not sure if changing to 9-digit will cause compatibility issue with existing redis client libraries. I know some client libraries read and parse the module version number.

PR #50, opened 2025-03-22
Change unstable version to six digit max.:999999

roshkhatri:
This would allow for testing on unstable for being compatible with the latest ver in future.


joehu21 (approved):
LGTM

KarthikSubbarao (approved):

PR #52, opened 2025-03-28
Decoupling JSON python tests from Bloom. Now using valkey-test-framework

zackcam:
Removed all mentions of bloom where we were pulling the test framework in the build script. Now we are pulling from valkey-test-framework. Due to this there were some changes that needed to be made, debug_digest is no longer a method as that was just a wrapper on running debug digest. We also now explicitly create the servers and clients not in the testframework as this allows more control for the developers. 
This pr updated all the above to be compatible with the new test framework


roshkhatri (approved):
LGTM, Thank you so much

PR #57, opened 2025-04-01
Valkey-json release 1.0.0 GA

roshkhatri:


src/commands/json.mset.json
```diff
@@ -0,0 +1,36 @@
+{
+    "JSON.MSET": {
+        "summary": "Set multiple JSON values at the path to multiple keys.",
+        "complexity": "O(N) where N is the number of keys",
+        "group": "json",
+        "module_since": "1.0.0",
+        "arity": -4,
+        "acl_categories": [
+            "WRITE",
+            "SLOW",
```
ranshid, line 10: any reason this is considered "slow"? for example HMSET (in Valkey) is not considered "slow"

src/commands/json.set.json
```diff
@@ -0,0 +1,34 @@
+{
+    "JSON.SET": {
+        "summary": "Set JSON values at the specified path.",
+        "complexity": "O(N) where N is the number of json values matched by the path.",
+        "group": "json",
+        "module_since": "1.0.0",
+        "arity": 4,
+        "acl_categories": [
+            "WRITE",
+            "SLOW",
```
ranshid, line 10: why is that a "slow" command?

roshkhatri: Set involves deserialization and hence considered as slow as the main string is large.

src/commands/json.mset.json
joehu21: set and mset involves parsing JSON string. If JSON string is large, parsing is slow.

joehu21 (approved):
LGTM

PR #58, opened 2025-04-01
trim debug command summary

roshkhatri:
Diff of the older version of repidjson and the new updated sha: https://github.com/Tencent/rapidjson/compare/0d4517f15a8d7167ba9ae67f3f22a559ca841e3b..ebd87cb468fb4cb060b37e579718c4a4125416c1

`CMake version 4.0 is the first version to remove compatibility for CMake versions older than 3.5. While CMake 3.27 issued a deprecation warning for compatibility with versions older than 3.5, the actual removal of support occurred in version 4.0. Projects using older versions of CMake will need to be updated to at least version 3.5 or later to work with CMake 4.0 and newer.`


joehu21 (approved):
LGTM

PR #67, opened 2025-06-19
Improve build script

### LINKED ISSUE #62: Improve on build script and build options.
Currently there is no option to run the build.sh script for `release` and mentions threads at the build script level to specify the threads used for builds.

We would want to add option to build release only, where
   1. we only build the module and not build valkey server.
   1. not build the tests for release options.

roshkhatri:
Adds these options and build efficiently
```
Usage: build.sh [--release] [--unit] [--integration] [--clean]
    --help | -h               Print this help message and exit.
    --release                 Builds the release configuration.
    --unit                    Builds the unit tests configuration.
    --integration             Builds the integration tests configuration.
    --clean                   Cleans the build artifacts.
Example usage:
    # Build the release configuration,
    ./build.sh --release
    # Cleans the build artifacts,
    ./build.sh --clean
```

also updated the workflow to run all options.
fixes #62 

Refer the workflows that tests the script and also runs unit tests and integration tests for all versions

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,46 +3,72 @@ name: CI
 on: [push, pull_request]
 
 jobs:
-  build-ubuntu-latest:
+  release-build:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
         server_version: ['unstable', '8.0', '8.1']
     steps:
-    - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
-    - name: Set the server version for python integration tests
-      run: echo "SERVER_VERSION=${{ matrix.server_version }}" >> $GITHUB_ENV
-    - name: Set up Python
-      uses: actions/setup-python@3542bca2639a428e1796aaa6a2ffef0c0f575566 #v3.1.4
-      with:
-        python-version: '3.9'
-    - name: Install dependencies
-      run: |
-        python -m pip install --upgrade pip
-        pip install -r requirements.txt
-    - name: Build and Run tests. 
-      run: ./build.sh
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      - name: Set the server version
+        run: echo "SERVER_VERSION=${{ matrix.server_version }}" >> $GITHUB_ENV
+      - name: Build release
+        run: ./build.sh --release
 
-  build-asan-ubuntu-latest:
+  unit-tests:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
         server_version: ['unstable', '8.0', '8.1']
     steps:
-    - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
-    - name: Set the server version for python integration tests
-      run: echo "SERVER_VERSION=${{ matrix.server_version }}" >> $GITHUB_ENV
-    - name: Set up Python
-      uses: actions/setup-python@3542bca2639a428e1796aaa6a2ffef0c0f575566 #v3.1.4
-      with:
-        python-version: '3.9'
-    - name: Install dependencies
-      run: |
-        python -m pip install --upgrade pip
-        pip install -r requirements.txt
-    - name: Build and Run tests. 
-      run: |
-        export ASAN_BUILD=true
-        ./build.sh
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      - name: Set the server version
+        run: echo "SERVER_VERSION=${{ matrix.server_version }}" >> $GITHUB_ENV
+      - name: Run unit tests
+        run: ./build.sh --unit
+
+  integration-tests:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        server_version: ['unstable', '8.0', '8.1']
+    steps:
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -3,46 +3,72 @@ name: CI
...
+        run: echo "SERVER_VERSION=${{ matrix.server_version }}" >> $GITHUB_ENV
+      - name: Run unit tests
+        run: ./build.sh --unit
+
+  integration-tests:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        server_version: ['unstable', '8.0', '8.1']
```
sarthakaggarwal97, line 37: I guess we will have to amend this whenever there is a new release in Valkey

.github/workflows/ci.yml
```diff
@@ -3,46 +3,72 @@ name: CI
 on: [push, pull_request]
 
 jobs:
-  build-ubuntu-latest:
+  release-build:
```
sarthakaggarwal97:
```suggestion
  build-release:
```

CMakeLists.txt
```diff
@@ -31,7 +31,11 @@ endif()
 # Set the name of the JSON shared library
 set(JSON_MODULE_LIB json)
 
-# Define the Valkey directories
+option(RELEASE_BUILD "Build only valkey-json module" OFF)
```
sarthakaggarwal97:
```suggestion
option(BUILD_RELEASE "Build only valkey-json module" OFF)
```

build.sh
```diff
@@ -1,75 +1,145 @@
...
+RELEASE_BUILD=0
+CLEAN_BUILD=0
+
+## Parse command line argument
+while [[ $# -gt 0 ]]; 
+do
+    arg="$1"
+    case $arg in
+        --release)
+            RELEASE_BUILD=1
```
sarthakaggarwal97:
```suggestion
            BUILD_RELEASE=1
```

Probably needs to be changed in multiple places.

src/CMakeLists.txt
```diff
@@ -38,3 +39,4 @@ target_sources(${OBJECT_TARGET}
 )
 
 add_library(${JSON_MODULE_LIB} SHARED $<TARGET_OBJECTS:${OBJECT_TARGET}>)
+add_dependencies(${OBJECT_TARGET} valkey)
```
sarthakaggarwal97, line 42: do we need to add this dependency everytime? I am thinking we might need it only for integration tests?

README.md
```diff
@@ -4,73 +4,49 @@ Valkey-json is a Valkey module written in C++ that provides native JSON (JavaScr
...
-```
-
-#### To build just the module
+To build only valkey-json module, use:
 ```text
-mkdir build
-cd build
-cmake ..
-make
+./build.sh --release
```
sarthakaggarwal97: I am wondering if default behavior should just be to run without tests. And there could be an additional parameter to run the tests. Something like Valkey, where we use `make` to build and `make test` to run the tests and build.

CMakeLists.txt
```diff
@@ -90,41 +106,42 @@ ExternalProject_Add_Step(
...
+    set(VALKEY_TEST_FRAMEWORK_COMMIT "9fb28b74efd122324db00498a2fde6e5d281c90f" CACHE STRING "Valkey-test-framework commit to use")
+
+    # Set the download directory for Valkey-test-framework
+    set(VALKEY_TEST_FRAMEWORK_DOWNLOAD_DIR "${CMAKE_CURRENT_BINARY_DIR}/_deps/valkey-test-framework-src")
+
+    ExternalProject_Add(
+        valkey-test-framework
+        GIT_REPOSITORY https://github.com/valkey-io/valkey-test-framework.git
+        GIT_TAG ${VALKEY_TEST_FRAMEWORK_COMMIT}
+        GIT_SHALLOW FALSE
```
sarthakaggarwal97, line 118: do we need this to be true? imo we should not need to full history

.github/workflows/ci.yml
roshkhatri: Yes, I dont see an easier way to get all the supported versions as of now

CMakeLists.txt
roshkhatri: So we are not pulling the latest commit and if we make it shallow it is currently failing to clone it at the commit.
This is only for testing so it will have effect on the actual build/ build size

README.md
roshkhatri: Makes sense, will make the nec. changes

CMakeLists.txt
sarthakaggarwal97: > it will have effect on the actual build/ build size

do you mean it will not have an effect?

roshkhatri: Yes, sorry. It will not affect

src/CMakeLists.txt
roshkhatri: We would need these as we are copying the header file after cloning valkey

build.sh
```diff
@@ -1,75 +1,145 @@
...
+    case $arg in
+        --release)
+            BUILD_RELEASE=1
+            RUN_UNIT=0
+            RUN_INTEGRATION=0
+            ;;
+        --unit)
+            RUN_UNIT=1
+            RUN_INTEGRATION=0
+            BUILD_RELEASE=0
```
sarthakaggarwal97, line 46: does `BUILD_RELEASE` need to be 0? I assume that to run the test, we will be building the project as well

sarthakaggarwal97 (approved):
LGTM! thanks for making this change

build.sh
roshkhatri: yes I hsve kept it to be a separate build in the CMAKELIST

PR #69, opened 2025-06-25
Support building on Clang / Apple

eifrah-aws:
- Enabled `compile_commands.json` in `CMake` for better code completion
- Do not call "std::move" on temporary variables
- Fixed formatting errors (`printf` wrong identifiers)
- Fixed invalid struct definitions
- Fixed unused parameter values (by adding: `[[maybe_unused]]`)
- Fixed: multiple symbols link error on clang by decorating Valkey methods as `attribute(weak)`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,8 +27,27 @@ endif()
 project(ValkeyJSONModule VERSION 99.99.99 LANGUAGES C CXX) # Version is 99.99.99 for unstable branch
 
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-mismatched-tags -Wno-format")
+
+# Clang related adjustments
+set(COMPILER_IS_CLANG OFF)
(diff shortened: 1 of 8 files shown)


joehu21 (commented):
I'm seeing compiler errors like `/Users/jowhuw/github/joehu21-valkey-json/src/json/util.h:13:10: fatal error: 'cmath' file not found` on my Macbook. I think we need to make an additional change. Add the following at the top of CMakeLists.txt:
```
if (APPLE)
    # Force clang compiler on macOS
    find_program(CLANGPP "clang++")
    find_program(CLANG "clang")
    if (CLANG AND CLANGPP)
        message(STATUS "Found ${CLANGPP}, ${CLANG}")
        set(CMAKE_CXX_COMPILER ${CLANGPP})
        set(CMAKE_C_COMPILER ${CLANG})
    endif ()
endif ()
```

Then, it compiles on my Macbook.

joehu21 (approved):
LGTM

PR #70, opened 2025-08-04
Modify JSON Tests to allow connection to external server.

Nikhil-Manglore:
This change adds the ability to run tests against an external Valkey server instead of always starting a new one. It coincides with the changes made in the Valkey-Test-Framework (https://github.com/valkey-io/valkey-test-framework/pull/7)

External server mode: If we set the environment variable, `VALKEY_EXTERNAL_SERVER=true , when we run the bloom pytests we will connect to an existing Valkey server at the specified host and port.

Local server mode: If not set or false, we will continue with the original behavior of starting a new Valkey server with the module loaded


roshkhatri (approved):
LGTM

PR #71, opened 2025-08-05
Implement SharedAPI interface that mimics JSON.GET for Search

allenss-amazon:


yairgott (commented):
The name that I'm using now for the planned hash API is: [VM_HashSetValueView](https://github.com/valkey-io/valkey/compare/unstable...yairgott:valkey:hash-view#diff-6109c354d7e009093f811238069b581bcb9bdbfc638d7d089814031776801632R5238)

FWIW, it might be hard to follow as this PR is based on Ran's hash TTL PR.

I think that `View` makes better sense to me than `Shared`. WDYT?

src/json/shared_api.h
```diff
@@ -0,0 +1,31 @@
...
+//
+// Fetch JSON text associated with "path".
+//
+// Errors:
+//
+//   If the key isn't JSON, then you'll get REDISMODULE_ERR for the return.
+//   If the key is JSON, but the path doesn't identify anything, then you'll get a nullptr for the result.
+//   Otherwise you get the JSON text that matches the path.
+//
+int SharedJSON_Get(ValkeyModuleKey *key, const char *path, ValkeyModuleString **result);
```
yairgott, line 24: As discussed, rename to `JSON_GetValue`

src/json/shared_api.cc
```diff
@@ -0,0 +1,30 @@
...
+#include "json/json.h"
+#include "json/dom.h"
+#include "json/selector.h"
+
+void SharedAPI_Register(ValkeyModuleCtx *ctx) {
+    ValkeyModule_ExportSharedAPI(ctx, "SharedJSON_Get", (void *)SharedJSON_Get);
+}
+
+int SharedJSON_Get(ValkeyModuleKey *key, const char *path, ValkeyModuleString **result) {
+    *result = nullptr;
```
yairgott: Avoid assigning to `*result` on error.

src/json/shared_api.cc
```diff
@@ -0,0 +1,30 @@
...
+    if (verify_open_doc_key(key) != JSONUTIL_SUCCESS) {
+        return VALKEYMODULE_ERR;
+    }
+    // Fetch the document from the key
+    JDocument *doc = static_cast<JDocument *>(ValkeyModule_ModuleTypeGetValue(key));
+    PrintFormat format;
+    ReplyBuffer output;
+    return dom_get_value_as_str(doc, path, &format, output);
```
yairgott: return only on error.

src/json/shared_api.cc
```diff
@@ -0,0 +1,30 @@
...
+ *
+ */
+
+#include "shared_api.h"
+#include "json/json.h"
+#include "json/dom.h"
+#include "json/selector.h"
+
+void SharedAPI_Register(ValkeyModuleCtx *ctx) {
+    ValkeyModule_ExportSharedAPI(ctx, "SharedJSON_Get", (void *)SharedJSON_Get);
```
yairgott: Missing error handling, perhaps an assert.

yairgott:
> The name that I'm using now for the planned hash API is: [VM_HashSetValueView](https://github.com/valkey-io/valkey/compare/unstable...yairgott:valkey:hash-view#diff-6109c354d7e009093f811238069b581bcb9bdbfc638d7d089814031776801632R5238)
> 
> FWIW, it might be hard to follow as this PR is based on Ran's hash TTL PR.
> 
> I think that `View` makes better sense to me than `Shared`. WDYT?

Lets discuss the memory sharing additional interfaces separately.

yairgott (approved):
Approving. Please address the open comments before submitting.

src/json/shared_api.cc
allenss-amazon: Why? Seems safer to me that result is unconditionally written.

src/json/shared_api.h
allenss-amazon: done.

src/json/shared_api.cc
allenss-amazon: oops

src/json/shared_api.cc
allenss-amazon: ack.

src/json/shared_api.cc
yairgott: > Why? Seems safer to me that result is unconditionally written.

The general guidance is no side effects on failure. I agree that in this particular case, assigning a `nullptr` unlikely would lead to issues but also there is no value in doing so as the caller should rely on the return code for errors.

src/json/shared_api.cc
```diff
@@ -0,0 +1,34 @@
...
+#include "json/dom.h"
+#include "json/selector.h"
+
+void SharedAPI_Register(ValkeyModuleCtx *ctx) {
+    if(ValkeyModule_ExportSharedAPI(ctx, "JSON_GetValue", (void *)SharedJSON_Get) != VALKEYMODULE_OK) {
+        ValkeyModule_Assert(false);
+    }
+}
+
+int SharedJSON_Get(ValkeyModuleKey *key, const char *path, ValkeyModuleString **result) {
```
yairgott, line 19: Proposed implementation below. The main difference is that it requires to use `rapidjson::StringBuffer` rather than `ReplyBuffer`.

```
int SharedJSON_Get(ValkeyModuleKey *key, const char *path, ValkeyModuleString **result) {
    if (verify_open_doc_key(key) != JSONUTIL_SUCCESS) {
        return VALKEYMODULE_ERR;
    }
    // Fetch the document from the key
    JDocument *doc = static_cast<JDocument *>(ValkeyModule_ModuleTypeGetValue(key));
    rapidjson::StringBuffer buf;
    if (dom_get_value_as_str(doc, path, nullptr, buf) == JSONUTIL_SUCCESS) {
        auto str = ValkeyModule_CreateString(nullptr, buf.GetString(), buf.GetLength());
        *result = str;
        return VALKEYMODULE_OK;
    }
    return VALKEYMODULE_ERR;
}
```

yairgott (approved):

src/json/shared_api.h
```diff
@@ -0,0 +1,32 @@
...
+#include "./include/valkeymodule.h"
+
+#include <stddef.h>
+
+//
+// Fetch JSON text associated with "path".
+//
+// Errors:
+//
+//   If the key isn't JSON, then you'll get REDISMODULE_ERR for the return.
```
roshkhatri:
```suggestion
//   If the key isn't JSON, then you'll get VALKEYMODULE_ERR for the return.
```

roshkhatri (commented):
LGTM!
There are just a few comments
- just one nitpick `REDISMODULE_ERR` -> `VALKEYMODULE_ERR`
- is there a way we could some tests for this in JSON??
- Can we also rebase on this branch? The reason for this ask is that the git history for the other commits would change to this commit.

roshkhatri (approved):
LGTM

Issue #10, opened 2024-11-29
Avoid downloading dependencies during the build


madolson:
I would like to see some way to not download dependencies during the build, at least optionally.

Offline builds are common to guarantee that a project doesn't have any online dependencies that have not been verified in advance. Supply chain attacks is the main concern. License/legal problems is another.

I suggest we add some environmant variable to provide the dependencies using a local paths and CMake can use these instead of downloading them if they are provided. It can be added in the future though.

_Originally posted by @zuiderkwast in https://github.com/valkey-io/valkeyJSON/pull/1#discussion_r1858979053_

PR #73, opened 2025-08-06
Simplify offline build

remicollet:
For distribution package  (ex: RPM) the build is expected to be run offline.

- add RAPIDJSON_SOURCE_DIR to avoid RapidJSON download

- add VALKEY_INCLUDE_DIR to avoid valkey download

ex: `-DVALKEY_INCLUDE_DIR=/usr/include` as the `valkey-devel `package provides this header.


remicollet:
For your information, I'm working on RPM packaging of valkey, and its modules

My work is available in my private repository, and may be integrated in Fedora official repository later

See https://blog.remirepo.net/post/2025/08/01/Valkey-version-8.1
And https://git.remirepo.net/cgit/rpms/valkey/valkey-json.git/tree/

src/CMakeLists.txt
```diff
@@ -39,4 +39,7 @@ target_sources(${OBJECT_TARGET}
 )
 
 add_library(${JSON_MODULE_LIB} SHARED $<TARGET_OBJECTS:${OBJECT_TARGET}>)
-add_dependencies(${OBJECT_TARGET} valkey)
+
+if(ENABLE_UNIT_TESTS OR ENABLE_INTEGRATION_TESTS)
+	add_dependencies(${OBJECT_TARGET} valkey)
+endif()
```
remicollet, line 46: Is this dependency really needed ?

It blocks offline build (without valley tree)
build with `BUILD_RELEASE`, without `ENABLE_UNIT_TESTS` and without `ENABLE_INTEGRATION_TESTS` works

So I make it conditional

roshkhatri:
I think this would also resolve: https://github.com/valkey-io/valkey-json/issues/10

Can you also please sign-off the commits?

src/CMakeLists.txt
roshkhatri: the `BUILD_RELEASE` would require `./include/valkeymodule.h` and thats why we would need the valkey
But I think would be resolved if we provide the path-to-valkey

We would also need to make changes to the CI to include `VALKEY_INCLUDE_DIR`

roshkhatri:
We would also need to checkout valkey and add the path to the tests in order to run them as an offline build.

roshkhatri (approved):
LGTM, added the path for build release in ci.yml

Issue #75, opened 2025-08-11
[Question] Has valkey-json considered adding .clang-format and .clang-tidy files?


cxljs:
Hello, has valkey-json considered adding .clang-format and .clang-tidy files? For example, using valkey-search's .clang-format, or @roshkhatri 's?

sarthakaggarwal97:
I believe we should add these to the repo. @cxljs do you want to take a look or I can take a stab at it.

cxljs:
Sounds good! I can take a stab at it and submit a PR for review. I want to use valkey-search's .clang-format and .clang-tidy. Do you think that's okay?

Nikhil-Manglore:
Yeah I think that sounds like a good idea

cxljs:
Got it!

sarthakaggarwal97:
Thanks @cxljs for taking this up!

cxljs:
I'm sorry I've been very busy recently and couldn't resolve this issue quickly. I've already submitted a PR.

PR #78, opened 2025-08-23
Add .clang-format and .clang-tidy

cxljs:
refs: #75

### DIFF
--- a/src/json/alloc.h
+++ b/src/json/alloc.h
@@ -16,7 +16,7 @@
 #ifndef VALKEYJSONMODULE_ALLOC_H_
 #define VALKEYJSONMODULE_ALLOC_H_
 
-#include <stddef.h>
+#include <cstddef>
 
 #include "json/memory.h"
 
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,3 @@
+BasedOnStyle: Google
+IndentWidth: 4
+ColumnLimit: 120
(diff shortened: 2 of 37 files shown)


roshkhatri:
Hey, Thank you for working on this, Can you also add a workflow for this check on very push and pull req?
Similar implementation can be found here in valkey: https://github.com/valkey-io/valkey/blob/unstable/.github/workflows/clang-format.yml:


Do we also want to use this: https://github.com/valkey-io/valkey-json/blob/unstable/src/json/CPPLINT.cfg?

cxljs:
> Can you also add a workflow for this check on very push and pull req?

Yes, I will finish it later.

> Do we also want to use this: https://github.com/valkey-io/valkey-json/blob/unstable/src/json/CPPLINT.cfg?

I was concerned that you might need the cpplint.cfg file, so I kept it. If you don't need it, I'll remove it later.

roshkhatri:
Okay will wait and review again when wfs are added.

src/json/alloc.h
```diff
@@ -16,7 +16,7 @@
 #ifndef VALKEYJSONMODULE_ALLOC_H_
 #define VALKEYJSONMODULE_ALLOC_H_
 
-#include <stddef.h>
+#include <cstddef>
```
sarthakaggarwal97, line 19: 👍

sarthakaggarwal97 (commented):
It took me some time to get back to this, sorry about that. The changes looks mostly good to me. @cxljs are you looking into adding a new workflow?

Nikhil-Manglore:
Tagging @cxljs in case they missed the previous notifications 😄

roshkhatri:
I think we can close this as it has been inactive for a long time. If req, please open this again

PR #95, opened 2025-12-30
feat: JSON.MERGE plus unit and integration tests

### LINKED ISSUE #74: Support JSON.MERGE
Can you support JSON.MERGE, just like RedisJSON?

JSON.SET directly overwrites the subtree with the new value, but does not perform the Merge operation.

Now, Merge a JSON subtree, requiring:
1. JSON.GET, 
2. Write code Merge
3. JSON.SET

But JSON.MERGE can do it .

Another important reason is that Redis JSON uses the RUST language and valkey-json uses the C++ language. 

This makes valkey-json the best choice in scenarios where Rust is not the development language.

patrickn2:
JSON.MERGE Command
Followed instructions here https://redis.io/docs/latest/commands/json.merge/

Closes #74.


roshkhatri:
Can you please add --sign off to the commits?: https://github.com/valkey-io/valkey-json/pull/95/checks?check_run_id=59176701849

patrickn2:
> Can you please add --sign off to the commits?: https://github.com/valkey-io/valkey-json/pull/95/checks?check_run_id=59176701849

I guess I made a mess trying to fix it
Could you take a look?

roshkhatri:
> > Can you please add --sign off to the commits?: https://github.com/valkey-io/valkey-json/pull/95/checks?check_run_id=59176701849
> 
> I guess I made a mess trying to fix it Could you take a look?

Yeah, its okay , you can just run these commands and it should be fine:
```
git rebase HEAD~3 --signoff
git push --force-with-lease origin feat--JSON.MERGE-Command
```

patrickn2:
> > > Can you please add --sign off to the commits?: https://github.com/valkey-io/valkey-json/pull/95/checks?check_run_id=59176701849
> > 
> > 
> > I guess I made a mess trying to fix it Could you take a look?
> 
> Yeah, its okay , you can just run these commands and it should be fine:
> 
> ```
> git rebase HEAD~3 --signoff
> git push --force-with-lease origin feat--JSON.MERGE-Command
> ```

Done

src/json/dom.cc
```diff
@@ -208,6 +210,181 @@ JsonUtilCode dom_set_value(ValkeyModuleCtx *ctx, JDocument *doc, const char *jso
...
+ * - Recursive calls process each node exactly once
+ * 
+ * Merge behavior:
+ * - If both are objects: merge their members recursively
+ * - Otherwise: return a copy of the new value
+ * - Null values delete keys (existing) or prevent addition (new)
+ */
+JValue merge_values(const JValue &existing, const JValue &new_val, RapidJsonAllocator &alloc, int depth) {
+    // Safety: prevent infinite recursion (max depth 100)
+    if (depth > 100) {
```
roshkhatri: Can this be dynamic and set to `max-path-limit` config?

src/json/json.cc
```diff
@@ -735,6 +735,94 @@ int Command_JsonSet(ValkeyModuleCtx *ctx, ValkeyModuleString **argv, int argc) {
...
+            if (rc != JSONUTIL_SUCCESS) {
+                return ValkeyModule_ReplyWithError(ctx, jsonutil_code_to_message(rc));
+            }
+
+            // end tracking memory
+            END_TRACKING_MEMORY(ctx, "JSON.MERGE", doc, orig_doc_size, begin_val)
+
+            // update stats
+            size_t new_doc_size = dom_get_doc_size(doc);
+            jsonstats_update_stats_on_update(doc, orig_doc_size, new_doc_size, json_len);
```
roshkhatri: these two conditions are mostly the same, is there a way we can merge the two?

src/json/dom.cc
```diff
@@ -208,6 +210,181 @@ JsonUtilCode dom_set_value(ValkeyModuleCtx *ctx, JDocument *doc, const char *jso
...
+        return JValue(new_val, alloc);
+    }
+    
+    if (!existing.IsObject() || !new_val.IsObject()) {
+        // Not both objects, return copy of new value
+        return JValue(new_val, alloc);
+    }
+
```
roshkhatri, line 245: Can we add early exits here like:
```
if (existing.ObjectEmpty()) {
    // Nothing to merge into – just return a deep copy of new_val
    return JValue(new_val, alloc);
}

if (new_val.ObjectEmpty()) {
    // Patch is empty – return a deep copy of existing
    return JValue(existing, alloc);
}
```

CMakeLists.txt
roshkhatri, line 140: I don't think these changes are for this feature
Can you please rebase with the unstable branch so the PR only has changes related to this feature?

build.sh
roshkhatri, line 111: I don't think these changes are for this feature
Can you please rebase with the unstable branch so the PR only has changes related to this feature?

tst/unit/dom_test.cc
```diff
@@ -446,6 +446,270 @@ TEST_F(DomTest, testGetObject) {
...
+    // Key 'b' should not be added (null on non-existing key)
+    EXPECT_TRUE(strstr(result, "\"b\"") == nullptr);
+    // Key 'c' should be added
+    EXPECT_TRUE(strstr(result, "\"c\":3") != nullptr);
+    // Original key 'a' should remain
+    EXPECT_TRUE(strstr(result, "\"a\":1") != nullptr);
+    
+    dom_free_doc(doc);
+}
+
```
roshkhatri, line 713: Can we also add unit tests for array replacements with array, scalar or objects?
Also for empty objects, which would be a no-op

tst/integration/test_json_basic.py
```diff
@@ -4195,11 +4195,241 @@ def test_insert_update_delete_mode(self):
...
+        assert b'[4,5]' == client.execute_command('JSON.GET', key, '.arr')
+        
+        # Verify object merged (x preserved, y added)
+        obj_result = client.execute_command('JSON.GET', key, '.obj')
+        assert b'"x":1' in obj_result or b'"x": 1' in obj_result
+        assert b'"y":2' in obj_result or b'"y": 2' in obj_result
+        
+        # Verify new key added
+        assert b'"added"' == client.execute_command('JSON.GET', key, '.new')
+
```
roshkhatri, line 4515: Can we add some integration tests for
1. wild cards/ multiple targets, 
2. mixed for updates and inserts, like first user with `.profile` and second user without `.profile`
3. Non-root path on new key throws error
4. merge {} results in unchanged doc

tst/integration/conftest.py
```diff
@@ -0,0 +1,128 @@
+"""
+Local conftest.py to apply fixes to the test infrastructure.
```
roshkhatri, line 2: Do we need to upstream this to https://github.com/valkey-io/valkey-test-framework?

docs/JSON.MERGE_COMPLEXITY.md
```diff
@@ -0,0 +1,115 @@
+# JSON.MERGE Time Complexity
```
roshkhatri: We can removed this file before we merge this

docs/JSON.MERGE_SUMMARY.md
```diff
@@ -0,0 +1,173 @@
+# JSON.MERGE Command - Complete Implementation Summary
```
roshkhatri: We can removed this file before we merge this

docs/JSON.MERGE_SUMMARY.md
```diff
@@ -0,0 +1,173 @@
...
+
+- Uses RapidJSON's custom allocator integrated with Valkey's memory tracking
+- Properly tracks memory with `jsonstats_begin_track_mem()` and `END_TRACKING_MEMORY`
+- Updates document size statistics after merge
+- Handles OOM conditions with `deny-oom` flag
+
+## Conclusion
+
+The JSON.MERGE command is fully implemented, tested, and optimized with:
+- ✅ **O(M+N)** time complexity for efficient merging
```
roshkhatri: Can we please remove the emojis?

roshkhatri (commented):
Great PR, just some suggestions and may need to add more tests

src/json/dom.cc
patrickn2: Done

src/json/json.cc
patrickn2: Done

src/json/dom.cc
patrickn2: Done

tst/unit/dom_test.cc
patrickn2: Done

tst/integration/test_json_basic.py
patrickn2: Done

docs/JSON.MERGE_COMPLEXITY.md
patrickn2: Done

docs/JSON.MERGE_SUMMARY.md
patrickn2: Done

docs/JSON.MERGE_SUMMARY.md
patrickn2: File was removed

CMakeLists.txt
patrickn2: done

build.sh
patrickn2: Done

tst/integration/conftest.py
patrickn2: Not sure
At least here in my machine, integration tests were flaky, freezing and other weird behaviors
If in yours is running fine we can ignore it.

roshkhatri:
The failing CI would be fixed once we merge the latest unstable into this branch

tst/integration/conftest.py
patrickn2: @roshkhatri do I have to do anything else?

roshkhatri:
I am not sure though why these are failing: https://github.com/valkey-io/valkey-json/actions/runs/20278225578/job/58232842208
here the cloning into valkey is working but on this PR there seems to be some issue

tst/integration/conftest.py
roshkhatri: I want to run this by @KarthikSubbarao to see if we can add these fixes to the test infrastructure rather than keeping these local

patrickn2: ok thanks
Let me know if you need anything else.

src/CMakeLists.txt
```diff
@@ -40,4 +40,7 @@ target_sources(${OBJECT_TARGET}
 )
 
 add_library(${JSON_MODULE_LIB} SHARED $<TARGET_OBJECTS:${OBJECT_TARGET}>)
-add_dependencies(${OBJECT_TARGET} valkey)
+
```
roshkhatri: I think this is causing the CI to fail and I dont think this belongs to this feature

patrickn2: Omg, I replaced that.
let me check

patrickn2: Done, this was a bad merge conflict that I had last commit.
Sorry

src/json/dom.cc
```diff
@@ -208,6 +210,192 @@ JsonUtilCode dom_set_value(ValkeyModuleCtx *ctx, JDocument *doc, const char *jso
...
+        return JValue(new_val, alloc);
+    }
+    
+    if (new_val.ObjectEmpty()) {
+        // Patch is empty – return a deep copy of existing
+        return JValue(existing, alloc);
+    }
+
+    // Safety: prevent infinite recursion (max depth from config)
+    if (depth > static_cast<int>(json_get_max_path_limit())) {
```
sarthakaggarwal97: Is it possible that this limit is breached while merging two deeply nested json? What would be the expected behaviour?

Ideally, this is right that we use the same limit, but we should document this.

src/json/dom.cc
```diff
@@ -208,6 +210,192 @@ JsonUtilCode dom_set_value(ValkeyModuleCtx *ctx, JDocument *doc, const char *jso
...
+        if (new_it != new_val.MemberEnd()) {
+            // Key exists in both
+            if (new_it->value.IsNull()) {
+                // Rule 1: Merging with null value deletes the key
+                // Don't add this key to merged object
+                continue;
+            } else if (it->value.IsObject() && new_it->value.IsObject()) {
+                // Both are objects - merge recursively
+                JValue val_result = merge_values(it->value, new_it->value, alloc, depth + 1);
+                JValue key_copy;
```
sarthakaggarwal97: can we reuse this variable across iterations?

src/json/dom.cc
```diff
@@ -208,6 +210,192 @@ JsonUtilCode dom_set_value(ValkeyModuleCtx *ctx, JDocument *doc, const char *jso
...
+ * - Second pass: O(N) - iterate new keys, O(1) check in merged
+ * - Recursive calls process each node exactly once
+ * 
+ * Merge behavior:
+ * - If both are objects: merge their members recursively
+ * - Otherwise: return a copy of the new value
+ * - Null values delete keys (existing) or prevent addition (new)
+ */
+JValue merge_values(const JValue &existing, const JValue &new_val, RapidJsonAllocator &alloc, int depth) {
+    if (existing.ObjectEmpty()) {
```
sarthakaggarwal97: Might be good to gate it with the check here?

```suggestion
    if (existing.IsObject() && existing.ObjectEmpty()) {
```

src/json/json.cc
```diff
@@ -2983,6 +3063,9 @@ extern "C" int ValkeyModule_OnLoad(ValkeyModuleCtx *ctx) {
     if (!set_command_info(ctx, "JSON.SET", -4, ks_read_write_update, 1, std::make_tuple(0, 1, 0))) {
         return VALKEYMODULE_ERR;
     }
+    if (!set_command_info(ctx, "JSON.MERGE", -4, ks_read_write_update, 1, std::make_tuple(0, 1, 0))) {
```
sarthakaggarwal97: `-4` implies >=4. In the code I think we check for exactly 4 arity

src/json/dom.cc
```diff
@@ -208,6 +210,192 @@ JsonUtilCode dom_set_value(ValkeyModuleCtx *ctx, JDocument *doc, const char *jso
...
+        if (has_updates) {
+            // If we had updates, we can't use the original selector's commit
+            // because it will try to process updates again and overwrite our merged values.
+            // Create a fresh selector that will only find inserts (paths that don't exist)
+            Selector insert_selector;
+            rc = insert_selector.prepareSetValues(root, json_path);
+            if (rc == JSONUTIL_SUCCESS && insert_selector.hasInserts()) {
+                // This selector should only have inserts since we already handled updates
+                // But to be safe, we'll only commit if it has no updates
+                if (!insert_selector.hasUpdates()) {
```
sarthakaggarwal97: can we add a test where we have both updates and inserts together?
```
{
  "users": [
    { "id": 1, "profile": { "email": "a@x.com" } },
    { "id": 2 }
  ]
}

JSON.MERGE key .users[*].profile {"phone":"123"}
```

sarthakaggarwal97 (commented):
Few comments from my end. Thank you for the change.

src/json/dom.cc
patrickn2: Adding some documentation about it in Readme.md

src/json/dom.cc
patrickn2: Done

src/json/json.cc
patrickn2: Done

src/json/dom.cc
patrickn2: The JSON.MERGE command was not prepared to do that. I had to make some changes to be possible to update and insert.
I also checked every edge case of the RFC https://datatracker.ietf.org/doc/html/rfc7396 
Had to add new tests and new implementations to compiles with it.
Thank you for your review 😁

sarthakaggarwal97:
@patrickn2 thanks for the fixes. Allow me some time to review again!

.github/workflows/ci.yml
sarthakaggarwal97, line 13: why is this removed?

patrickn2: Sorry, bad merge conflict that I had
gonna restore it

patrickn2: Done
Let me know if I messed something else 😅

mkmkme, line 11: This is unrelated to the PR topic. Please revert that.

mkmkme, line 13: Also for some reason some deletions. Could you please make sure your PR does not touch `.github/workflows/ci.yml` at all?

src/json/dom.cc
```diff
@@ -208,6 +210,171 @@ JsonUtilCode dom_set_value(ValkeyModuleCtx *ctx, JDocument *doc, const char *jso
...
+        return JValue(new_val, alloc);
+    }
+
+    if (new_val.ObjectEmpty()) {
+        return JValue(existing, alloc);
+    }
+
+    if (!existing.IsObject()) {
+        JValue empty_obj(rapidjson::kObjectType);
+        return merge_values(empty_obj, new_val, alloc, depth);
```
mkmkme, line 239: I might be wrong, but isn't this effectively `return JValue(new_val, alloc);`?
If that is so, can't we modify the earlier statement to be

```cpp
if (!new_val.IsObject() || !existing.IsObject()) {
    return JValue(new_val, alloc);
}
```

IMHO it also "fixes" the case when the objects are:

```js
// Old value
{
    "foo": ["bar", "baz"]
}

// New value
{
    "foo": {}
}
```

In the current implementation, the **old** value will be kept intact, because of your early check on line 233. This violates RFC 7396:
```
   define MergePatch(Target, Patch):
     if Patch is an Object:
       if Target is not an Object:
         Target = {} # Ignore the contents and set it to an empty Object
```

src/json/dom.cc
```diff
@@ -208,6 +210,171 @@ JsonUtilCode dom_set_value(ValkeyModuleCtx *ctx, JDocument *doc, const char *jso
...
+            JValue key_copy;
+            key_copy.SetString(key.c_str(), static_cast<rapidjson::SizeType>(key.length()), alloc);
+            merged.AddMember(key_copy, val_result, alloc);
+        }
+    }
+    
+    // Second pass: add keys that only exist in new_val (Target[Name] = MergePatch(undefined, Value))
+    JValue empty_obj(rapidjson::kObjectType);
+    for (auto it = new_val.MemberBegin(); it != new_val.MemberEnd(); ++it) {
+        std::string key(it->name.GetString(), it->name.GetStringLength());
```
mkmkme: No need to allocate `std::string` here. You can use

```cpp
if (!merged.HasMember(it->name.GetString())) {
```

and avoid having `std::string key` here completely. In `SetString`, I would rather use

```cpp
key_copy.SetString(it->name.GetString(), it->name.GetStringLength(), alloc);
```

src/json/dom.cc
```diff
@@ -208,6 +210,171 @@ JsonUtilCode dom_set_value(ValkeyModuleCtx *ctx, JDocument *doc, const char *jso
...
+        return merge_values(empty_obj, new_val, alloc, depth);
+    }
+    
+    // Build merged object from scratch
+    JValue merged(rapidjson::kObjectType);
+    
+    // First pass: process existing members
+    for (auto it = existing.MemberBegin(); it != existing.MemberEnd(); ++it) {
+        // Get key string and store in a std::string to ensure it persists
+        std::string key(it->name.GetString(), it->name.GetStringLength());
```
mkmkme: What exactly do you mean by "ensure it persists"? AFAICS we are inside the loop and it will persist as long as we are on the same iteration. IMHO there's no need to store it in `std::string`. Storing it in `std::string` brings two problems:
1. We have to go back and forth between RapidJSON API and `std::string` one
2. We are having an additional allocation which we could've avoided

src/json/dom.cc
```diff
@@ -208,6 +210,171 @@ JsonUtilCode dom_set_value(ValkeyModuleCtx *ctx, JDocument *doc, const char *jso
...
+    JValue &new_val = new_val_parser.GetJValue();
+    
+    // Handle updates (path exists) - merge values
+    bool has_updates = false;
+    auto &rs = selector.getUniqueResultSet();
+    if (!rs.empty()) {
+        has_updates = true;
+        for (auto &vInfo : rs) {
+            JValue *existing_val = vInfo.first;
+            typedef rapidjson::GenericPointer<RJValue, RapidJsonAllocator> JPointerType;
```
mkmkme:
```suggestion
            using JPointerType = rapidjson::GenericPointer<RJValue, RapidJsonAllocator>;
```

src/CMakeLists.txt
mkmkme, line 43: Why was this changed?

build.sh
mkmkme, line 130: Why the lines were removed?

CMakeLists.txt
mkmkme: Why was this file changed?

mkmkme (commented):
Hey! Here are some opinionated drive-by comments on the PR. Could you have a look please?

.github/workflows/ci.yml
roshkhatri: I think @patrickn2 you need to rebase onto unstable.

tst/integration/conftest.py
roshkhatri: I think we can keep this here

roshkhatri (commented):
I dont have more to add to what @sarthakaggarwal97 and @mkmkme 
I think just rebasing onto unstable should work

roshkhatri:
@patrickn2, if possible, can you please take a look at the feedback?
Valkey will be releasing 9.1 rc1 soon and we can release new version for valkey-json, so we can have valkey-json in the next bundle release.

.github/workflows/ci.yml
patrickn2: ok

patrickn2: Hey @mkmkme Thank you for your review
Could you check this change? I restored ci.yaml from here https://github.com/valkey-io/valkey-json/blob/unstable/.github/workflows/ci.yml but it looks different in the comparison

patrickn2: I already did it but it looks like I have changed the file.

If  you check line eleven in the comparison uses double quotes but if you go to the original file in the unstable branch it is using single quote

src/json/dom.cc
patrickn2: Fixed the issue but I just had to move some lines down

`    if (new_val.ObjectEmpty()) {
        return JValue(existing, alloc);
    }
`


I also added the test that you mentioned

src/json/dom.cc
patrickn2: HasMember(it->name.GetString()) is a problem. The pointer is not guaranteed to be null-terminated (or stays valid in all cases), so HasMember could misbehave when the full test suite ran and lead to wrong merge results.

What I did: I replaced `std::string` by `std::string_view` that has no heap allocation and used hasMember with the key

src/json/dom.cc
patrickn2: Fixed, did the same replacing by `std::string_view` and fixing other uses

Also removed the comments, no need them

src/CMakeLists.txt
patrickn2: mb, I was not being able to run integration tests, forgot to rollback

build.sh
patrickn2: Same as not being able to run tests
Rolling it back

CMakeLists.txt
patrickn2: Rolling it back

patrickn2:
I guess I fixed everything guys
Thank you for your review @mkmkme @roshkhatri

build.sh
```diff
@@ -109,6 +109,10 @@ fi
 
 CMAKE_FLAGS="$CMAKE_FLAGS -DENABLE_UNIT_TESTS=${ENABLE_UNIT_TESTS} -DENABLE_INTEGRATION_TESTS=${ENABLE_INTEGRATION_TESTS} -DBUILD_RELEASE=${ENABLE_BUILD_RELEASE}"
 
+if [ -n "${VALKEY_INCLUDE_DIR}" ]; then
+    CMAKE_FLAGS="$CMAKE_FLAGS -DVALKEY_INCLUDE_DIR=${VALKEY_INCLUDE_DIR}"
+fi
+
```
roshkhatri: I dont think this should be here. was removed with the revert commit it think.

roshkhatri:
There were some stray changes in the history, removed and pushed it, should be easier to review the PR now

build.sh
patrickn2: Done

src/commands/json.merge.json
```diff
@@ -0,0 +1,30 @@
+{
+    "JSON.MERGE": {
+        "summary": "Merge JSON values at the specified path. Creates a new key if it does not exist, or merges new values with existing values if the key already exists.",
+        "complexity": "O(M+N) when path evaluates to a single value, where M is the size of the original value and N is the size of the new value. O(P*(M_avg+N)) when path evaluates to multiple values, where P is the number of matched paths, M_avg is the average size of original values, and N is the size of the new value.",
+        "group": "json",
+        "module_since": "1.0.0",
```
roshkhatri: shouldnt this be 1.1.0?
```suggestion
        "module_since": "1.1.0",
```

patrickn2: Once it is a new feature, yes I guess so
can I commit the suggestion?

roshkhatri: Yes, you commit this

sarthakaggarwal97 (commented):
Apologies for getting to this much later than I thought I would. I was playing with this change with AI. I think this needs one more pass. 

I can reproduce two issues:

1. Recursive descent behaves differently from `JSON.SET` when the path matches both a parent and a child.

   ```text
   JSON.SET k . '{"a":{"a":1}}'
   JSON.MERGE k '$..a' '5'
   JSON.GET k
   ```

   This returns `{"a":{"a":5}}`. For the same path, `JSON.SET k '$..a' 5` gives `{"a":5}`. It looks like `JSON.MERGE` keeps applying child paths after the parent was already replaced.

2. The max document size check seems to miss multi-target merges. If the same patch is merged into two matched paths, `JSON.MERGE` can return `OK` even though the final document is over `json.max-document-size`.

sarthakaggarwal97:
We should plan to get this merged soon so we can release this command this year!

patrickn2:
Ok I fixed your issue @sarthakaggarwal97 
I also did a double check against the RFC and fixed other edge cases.
please let me know if you guys need any other fix

patrickn2:
Fixing Lint errors

patrickn2:
Lint Errors fixed

src/json/dom.cc
```diff
@@ -208,6 +210,195 @@ JsonUtilCode dom_set_value(ValkeyModuleCtx *ctx, JDocument *doc, const char *jso
...
+    for (auto &vInfo : rs) {
+        const jsn::string &path = vInfo.second;
+
+        bool covered_by_ancestor = false;
+        for (auto &other : rs) {
+            if (&other == &vInfo) continue;
+            if (path_is_ancestor(other.second, path)) {
+                covered_by_ancestor = true;
+                break;
+            }
```
roshkhatri, line 356: Here, The time complexity would be O(N^2) for is any other path an ancestor of this path. we can mention that. Also N would be bounded by the doc size limits.

Not a blocker, just putting it out there.

roshkhatri (approved):
Hey, Thank you for working on this, my concerns were also inline with Sarthak's and they seem to be resolved.

Thanks @sarthakaggarwal97 for reviewing this, was waiting for it.

Thanks @patrickn2 for working on this.

sarthakaggarwal97:
We should have a documentation update as well for this.

roshkhatri:
@patrickn2 Can you open a PR for this to the [valkey-io/valkey-doc](https://github.com/valkey-io/valkey-doc) repository?

patrickn2:
> @patrickn2 Can you open a PR for this to the [valkey-io/valkey-doc](https://github.com/valkey-io/valkey-doc) repository?

Oh, Sure I can, let me work on it, I didn't know that you guys had this repo

patrickn2:
> @patrickn2 Can you open a PR for this to the [valkey-io/valkey-doc](https://github.com/valkey-io/valkey-doc) repository?

Let me know if this is enough https://github.com/valkey-io/valkey-doc/pull/456

PR #98, opened 2026-04-22
fix: gate JSON.DEBUG KEYTABLE-CORRUPT behind json.debug-mode config

roshkhatri:
## Issue

`JSON.DEBUG KEYTABLE-CORRUPT <string>` was an unrecoverable memory-exhaustion DoS. Each call allocated a `KeyTable_Layout` via `makeHandle()` but never released it. Any user with `+@json` ACL could leak ~200MB/s, and FLUSHALL could not reclaim the memory — only a server restart did.

## Fix

Three-layer defense while preserving the command's original fault-injection purpose:

1. **Hidden, immutable module config `json.debug-mode`** (default `false`) this is also implemented by valkey-search and gates `KEYTABLE-CORRUPT`, `KEYTABLE-CLEAR`, and `KEYTABLE-DISTRIBUTION`. Must be set via `--json.debug-mode yes` at startup.
2. **`admin` command flag** on the three gated subcommands - requires `@admin` ACL category and gates. Also implemented by valkey-search.
3. **Recoverable fault injection** — `KEYTABLE-CORRUPT` still injects a stray handle (so `KEYTABLE-CHECK` can detect it), but the handle is now stored in a module-global registry. A new **`JSON.DEBUG KEYTABLE-CLEAR`** subcommand drains the registry without restarting the server.

Log line no longer includes the attacker-controlled input- only length and refcount.

Design mirrors [valkey-search](https://github.com/valkey-io/valkey-search)'s `search.debug-mode` pattern.

## Behavior

| Command | `debug-mode=no` (default) | `debug-mode=yes` |
|---------|---------------------------|-------------------|
| `KEYTABLE-CORRUPT` / `CLEAR` / `DISTRIBUTION` | `ERR unknown subcommand` | Works (requires `@admin`) |
| Other `JSON.DEBUG` subcommands | Unchanged | Unchanged |

## Tests

- `CORRUPT` → `KEYTABLE-CHECK` raises `Mismatch` error (proves injection works)
- 3× `CORRUPT` → `CLEAR` returns 3 → `KEYTABLE-CHECK` passes
- Gating tests verify all three return `unknown subcommand` without `debug-mode`
- `COMMAND INFO` arity test updated for `KEYTABLE-CLEAR`

### DIFF
--- a/src/json/json.cc
+++ b/src/json/json.cc
@@ -35,6 +35,7 @@
 
 #include <string>
 #include <cmath>
+#include <vector>
 
 /* In unstable branch the module version is always "999999". */
 #define MODULE_VERSION 999999
@@ -74,11 +75,15 @@ static size_t config_max_recursive_descent_tokens = DEFAULT_MAX_RECURSIVE_DESCEN
 #define DEFAULT_MAX_QUERY_STRING_SIZE (128 * 1024)  // 128KB
 static size_t config_max_query_string_size = DEFAULT_MAX_QUERY_STRING_SIZE;
 
+static int config_debug_mode = 0;
+
 #define DEFAULT_KEY_TABLE_SHARDS 32768
 #define DEFAULT_HASH_TABLE_MIN_SIZE 64
 KeyTable *keyTable = nullptr;
 rapidjson::HashTableFactors rapidjson::hashTableFactors;
 rapidjson::HashTableStats   rapidjson::hashTableStats;
+
+static std::vector<KeyTable_Handle> debug_corrupt_handles;
 
 extern size_t hash_function(const char *text, size_t length);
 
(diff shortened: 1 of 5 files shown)


roshkhatri:
@mkmkme can you please take a look

mkmkme:
>  @mkmkme can you please take a look

Hey! I'll try to have a look today or tomorrow

hanxizh9910 (approved):
LGTM, gating behind an immutable hidden config and @admin ACL is the right approach here.
Checked keytable.h and confirmed no double-free risk: destroyHandle clears the internal pointer and the destructor just asserts it's null. Tests look good.

mkmkme (approved):
LGTM

src/json/json.cc
```diff
@@ -74,12 +76,17 @@ static size_t config_max_recursive_descent_tokens = DEFAULT_MAX_RECURSIVE_DESCEN
...
 
+static int config_debug_mode = 0;
+
 #define DEFAULT_KEY_TABLE_SHARDS 32768
 #define DEFAULT_HASH_TABLE_MIN_SIZE 64
 KeyTable *keyTable = nullptr;
 rapidjson::HashTableFactors rapidjson::hashTableFactors;
 rapidjson::HashTableStats   rapidjson::hashTableStats;
 
+static std::mutex debug_corrupt_mutex;
```
allenss-amazon: Why need a mutex?

roshkhatri: added it defensively for correctness, but not required, i will remove it

allenss-amazon (approved):

PR #100, opened 2026-05-03
ci: Add cpplint workflow to run linter on PRs

### LINKED ISSUE #3: Add workflow/step to run the linter automatically on PRs
https://github.com/valkey-io/valkeyJSON/pull/1#discussion_r1859146316

smkher:
Add a GitHub Actions workflow that runs cpplint on C++ source files in src/json/ and tst/unit/ directories. The workflow triggers on push and pull_request events, using the existing CPPLINT.cfg configurations in each directory.

Closes #3

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -1,0 +1,35 @@
+name: Linter
+
+on:
+  push:
+  pull_request:
+
+concurrency:
+  group: linter-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  cpplint:
+    name: C++ Lint
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+
+      - name: Set up Python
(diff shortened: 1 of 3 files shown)


smkher:
@roshkhatri There were 108 pre-existing linting errors and the diff for fixing all of these would have been massive. To avoid the large diff in production code, I updated the CPPLINT.cfg filters in both directories to suppress those categories so the workflow passes on the current code. New code will still be checked against the remaining enabled rules. Let me know if this is fine.

.github/workflows/linter.yml
```diff
@@ -0,0 +1,35 @@
...
+          python-version: '3.9'
+
+      - name: Install cpplint
+        run: pip install cpplint
+
+      - name: Run cpplint on src/json
+        run: cpplint --recursive src/json/
+
+      - name: Run cpplint on tst/unit
+        run: cpplint --recursive tst/unit/
```
roshkhatri, line 35: We can call it recursively with multiple paths, if it should be fine `cpplint --recursive src/json/ tst/unit/`
I don't have a preference.

roshkhatri (approved):
Thank you for working on this!, The updated the CPPLINT.cfg filters are good.
Yes, the massive diff is something we don't want and its good that the new code will be checked too.

PR #102, opened 2026-06-03
Support running integration tests against an external Valkey server

enjoy-binbin:
In #70, we modified the Python code to support it, but some content was
missing.

- README: document the VALKEY_EXTERNAL_SERVER / VALKEY_HOST / VALKEY_PORT
  environment variables for running integration tests against an
  already-running Valkey server with module loaded.

- tst/integration/run.sh: drop the "Kill old running test" pkill block
  so it does not terminate the user-managed external valkey-server
  before tests start.

- CMakeLists.txt: bump VALKEY_TEST_FRAMEWORK_COMMIT from a pinned hash
  to "unstable" to pull in the external-server support that is not
  available in the previously pinned commit.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -137,7 +137,7 @@ if(ENABLE_INTEGRATION_TESTS)
             COMMENT "Copied valkey-server to destination directory"
         )
     # Define valkey-test-framework commit id
-    set(VALKEY_TEST_FRAMEWORK_COMMIT "9fb28b74efd122324db00498a2fde6e5d281c90f" CACHE STRING "Valkey-test-framework commit to use")
+    set(VALKEY_TEST_FRAMEWORK_COMMIT "unstable" CACHE STRING "Valkey-test-framework commit to use")
 
     # Set the download directory for Valkey-test-framework
     set(VALKEY_TEST_FRAMEWORK_DOWNLOAD_DIR "${CMAKE_CURRENT_BINARY_DIR}/_deps/valkey-test-framework-src")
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -137,7 +137,7 @@ if(ENABLE_INTEGRATION_TESTS)
             COMMENT "Copied valkey-server to destination directory"
         )
     # Define valkey-test-framework commit id
-    set(VALKEY_TEST_FRAMEWORK_COMMIT "9fb28b74efd122324db00498a2fde6e5d281c90f" CACHE STRING "Valkey-test-framework commit to use")
+    set(VALKEY_TEST_FRAMEWORK_COMMIT "unstable" CACHE STRING "Valkey-test-framework commit to use")
```
roshkhatri:
```suggestion
    set(VALKEY_TEST_FRAMEWORK_COMMIT "33cc62660ea94d49bd6ab27cb8c09b345fc5d577" CACHE STRING "Valkey-test-framework commit to use")
```
I think we should still pin it to the latest commit so we dont break tests in the future

roshkhatri (commented):
One small comment also,
Do you think we need a ci test for 
everything else is good

CMakeLists.txt
enjoy-binbin: Good call out.

enjoy-binbin:
> Do you think we need a ci test for

Yes, that would be great if we have one.

roshkhatri (approved):
LGTM

PR #103, opened 2026-06-03
Support reusing externally built valkey-server and valkeymodule.h for test

enjoy-binbin:
Add VALKEY_SERVER_PATH and VALKEY_MODULE_H_PATH options to skip
building Valkey from source for integration tests:

- VALKEY_SERVER_PATH: copy the provided valkey-server binary into
  the integration test directory instead of building it.
- VALKEY_MODULE_H_PATH: copy the provided valkeymodule.h instead of
  fetching it from the cloned Valkey repo, so the header version can
  match the binary version.
- When both are set (or VALKEY_MODULE_H_PATH alone for non-integration
  builds), the Valkey repo is no longer cloned at all, enabling fully
  offline builds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,12 +52,15 @@ export ASAN_BUILD=true
 To skip building Valkey from source and reuse an externally built `valkey-server` binary, set `VALKEY_SERVER_PATH` to its absolute path.
 The binary will be copied into the integration test directory; `valkeymodule.h` is still fetched from the Valkey repo for compiling the module:
 ```text
-SERVER_VERSION=unstable VALKEY_SERVER_PATH=/path/to/valkey-server ./build.sh --integration
+SERVER_VERSION=unstable \
+VALKEY_SERVER_PATH=/path/to/valkey-server \
+./build.sh --integration
 ```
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -49,6 +49,11 @@ To build the module with ASAN and run tests
 export ASAN_BUILD=true
 ./build.sh --integration
 ```
+To skip building Valkey from source and reuse an externally built `valkey-server` binary, set `VALKEY_SERVER_PATH` to its absolute path.
+The binary will be copied into the integration test directory; `valkeymodule.h` is still fetched from the Valkey repo for compiling the module:
+```text
+SERVER_VERSION=unstable VALKEY_SERVER_PATH=/path/to/valkey-server ./build.sh --integration
+```
```
roshkhatri, line 58: I think we can also add a path to valkeymodule.h? so you can match the version of valkey you built and the valkeymodule.h version? with this we would also be able to build it offilne?

enjoy-binbin: Yes, good call out, added valkeymodule.h

roshkhatri (approved):
LGTM

Issue #104, opened 2026-06-05
JSON.DEL with wildcard path crashes server (SIGSEGV) when object keys contain '~'


madolson:
### Description

A `JSON.DEL` using a multi-match path (`$.*`, `$..*`) against an object whose keys contain a tilde (`~`) crashes the server with a SIGSEGV (out-of-bounds read). Any authenticated client can trigger it with two commands and no special configuration.

### Steps to reproduce

```
JSON.SET k '$' '{"~":1,"~~":2}'
JSON.DEL k '$.*'
```

The server dies; the client sees `Error: Server closed the connection`.

Also reproduces via recursive descent:

```
JSON.SET t '$' '{"o":{"~":1,"~a":2}}'
JSON.DEL t '$..*'
```

**Minimal condition:** an object with >=2 keys each containing `~`, deleted via a path that selects more than one such member. With a single match the comparator is never invoked, so it does not crash.

### Crash output

```
=== VALKEY BUG REPORT START ===
crashed by signal: 11, si_code: 1
Accessing address: 0xfffffffffffffffc
...
libjson.so pathCompare::operator()(...)
  -> std::set<...,pathCompare>::insert
  -> _Rb_tree::_M_get_insert_unique_pos
```

### Root cause

`escape_member_name_for_json_pointer()` (`src/json/selector.cc:2426`) escapes `/` -> `~1` but not `~` -> `~0`, violating RFC 6901. Keys containing `~` therefore produce malformed JSON pointers.

In `Selector::deleteValues()` (`src/json/selector.cc:678-689`) these paths are inserted into a `jsn::set<jsn::string, pathCompare>`. The comparator `pathCompare::operator()` (`src/json/selector.cc:611`) does:

```cpp
JPointer ptr1 = JPointer(path1);
auto depth1 = ptr1.GetTokenCount();   // 0 for the invalid pointer
... tokenArray1[depth1 - 1] ...       // tokenArray[(size_t)-1] -> OOB read
```

`GetTokenCount()` returns 0 for the invalid pointer, so `tokenArray[depth - 1]` reads at offset `(size_t)-1`, faulting at `0xfffffffffffffffc`.

### Suggested fix

Escape `~` -> `~0` (before `/` -> `~1`) in `escape_member_name_for_json_pointer()`, and/or guard `pathCompare` against `GetTokenCount() == 0` / `ptr.HasError()`.

### Environment

- valkey-json commit `a932a1c`, `valkey/valkey-bundle:unstable`
[...]

PR #108, opened 2026-06-05
Fix for crash when key contains ~

zackcam:
Fix for issue: https://github.com/valkey-io/valkey-json/issues/104

Follow RFC 6901 where we set ~ as ~0 to distinguish a regular ~ compared to a ~ created from parsing a /. This fix will set ~ as ~0 for parsing. This will prevent crashes where we originally saw a ~ and expected it to be the start of parsing for a /. We have the order of tilde first so we don't double replace in the loop.

Tests: To make sure this doesnt get changed in the future we have the two scenarios listed in the issues as unit tests now. Without this change the tests would crash causing failures.

### DIFF
--- a/src/json/selector.cc
+++ b/src/json/selector.cc
@@ -616,6 +616,15 @@ struct pathCompare {
         JPointer ptr2 = JPointer(path2);
         size_t depth1 = ptr1.GetTokenCount();
         size_t depth2 = ptr2.GetTokenCount();
+
+        // Guard against invalid pointers. A malformed JSON pointer parses to zero tokens, and
+        // because depth is size_t, tokenArray[depth - 1] below would underflow to (size_t)-1 and
+        // read out of bounds. Fall back to a stable lexicographic ordering of the raw paths.
+        if (ptr1.HasError() || ptr2.HasError() || depth1 == 0 || depth2 == 0) {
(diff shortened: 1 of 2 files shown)


src/json/selector.cc
```diff
@@ -616,6 +616,7 @@ struct pathCompare {
         JPointer ptr2 = JPointer(path2);
         size_t depth1 = ptr1.GetTokenCount();
         size_t depth2 = ptr2.GetTokenCount();
+
         if (depth1 != depth2) return depth1 > depth2;
```
roshkhatri, line 629: I think we should  guard `pathCompare` against a zero-token pointer here.
Our depths are `size_t`  so if `GetTokenCount()` ever returns 0, `tokenArray[depth - 1]` would underflows to (size_t)-1 and go out of bounds

We are fixing for `~` but this will also guard us against crashes if `GetTokenCount` return a 0, wdyt?

roshkhatri (commented):
change looks goos, just one minor comment

src/json/selector.cc
zackcam: Makes sense it seems like a pretty free check as not intensive, will be a good guard incase there are any other gaps we have

roshkhatri (approved):

Issue #105, opened 2026-06-05
Out-of-range numeric values crash server (SIGABRT) via uncaught std::stod in GetDouble; replicable poison document


madolson:
### Description

Numeric values whose magnitude is outside the representable `double` range crash the server with a SIGABRT (uncaught C++ exception). This rapidjson fork stores doubles as their original decimal string and converts on read via `std::stod`, which throws `std::out_of_range` for subnormal-underflow and `>DBL_MAX` overflow values. The exception is uncaught, so `std::terminate` -> `abort()`.

This is effectively a **poison document**: `JSON.SET` stores the value losslessly without ever calling `GetDouble()`, so ingestion succeeds and the crash is deferred to any later read, comparison, or arithmetic. The poison value **replicates**, so a replica holding it crashes when it evaluates the value. `JSON.NUMINCRBY` is a write, so the crashing operation itself also propagates.

### Steps to reproduce

Underflow (subnormal operand):

```
JSON.SET k . '{"b":1.5}'
JSON.NUMINCRBY k .b 1e-308
```

Underflow (subnormal stored value, normal operand):

```
JSON.SET ks . '1e-308'
JSON.NUMINCRBY ks . 5
```

Overflow via a JSONPath filter literal (single read-only command, plain-digit form `1` followed by 309 zeros):

```
JSON.SET d '$' '{"n":5}'
JSON.GET d '$[?(@.n > 1000...0)]'    # literal = 10^309 > DBL_MAX
```

Each crashes the server (`Error: Server closed the connection`). Also reachable via `JSON.NUMMULTBY`, `EVAL_RO`, and inside `MULTI`/`EXEC`.

### Threshold

- Safe: magnitude `>= 2.2250738585072014e-308` and `<= 1e308`.
- Crashes: subnormal magnitudes (`1e-308`, `1e-310`, `4.9e-324`) and plain-digit literals `>= 10^309`.
- Note: exponent form `1e309` is **safe** (parsed to `inf`, not string-stored). Only plain-digit literals trigger the overflow path.

### Crash output

```
terminate called after throwing an instance of 'std::out_of_range'
  what():  stod
=== VALKEY BUG REPORT START ===
crashed by signal: 6, si_code: -6
...
_ZSt20__throw_out_of_rangePKc -> std::stod -> rapidjson::GenericValue::GetDouble
  -> dom_increment_by (src/json/dom.cc:430)
  -> Command_JsonNumIncrBy (src/json/json.cc:1026)
```

[...]

PR #109, opened 2026-06-05
Fixing out of range crash

zackcam:
Fix for: https://github.com/valkey-io/valkey-json/issues/105

This fix uses the suggested change of switching to strtod. As this will convert overflows to inf, and not just throw an error. This library still works on subnormal values as well. We do not need an extra check for ERANGE from what I could tell as we are covered by the inf check in `JSONUTIL_ADDITION_OVERFLOW`

Two new tests are created the one in dom checks for three different cases, normal + subnormal, subnormal + normal then an overflow check.

### DIFF
--- a/tst/unit/dom_test.cc
+++ b/tst/unit/dom_test.cc
@@ -777,6 +777,44 @@ TEST_F(DomTest, testNumIncrMultBy_string_value_overflow) {
     EXPECT_FALSE(isV2Path);
 }
 
+// Numbers outside the normal double range are kept in string form, and reading them via GetDouble()
+// must not crash. std::strtod instead returns a subnormal on underflow and +/-inf on overflow, so
+// underflow goes through as normal arithmetic and overflow is caught by the existing isinf check as
+// JSONUTIL_ADDITION_OVERFLOW.
+TEST_F(DomTest, testNumIncrBy_double_out_of_range) {
+    jsn::vector<double> res;
+    bool isV2Path;
+    JParser parser;
+
+    // Case 1: Stored value is normal, the increment 1e-308 is subnormal and is converted back via
+    // GetDouble() during the addition.
+    JsonUtilCode rc = dom_set_value(nullptr, doc1, ".foo", "1.5", false, false);
+    EXPECT_EQ(rc, JSONUTIL_SUCCESS);
+    rc = dom_increment_by(doc1, ".foo", &parser.Parse("1e-308", 6).GetJValue(), res, isV2Path);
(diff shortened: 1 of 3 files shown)


tst/unit/dom_test.cc
```diff
@@ -777,6 +777,44 @@ TEST_F(DomTest, testNumIncrMultBy_string_value_overflow) {
...
+
+    // Case 1: Stored value is normal, the increment 1e-308 is subnormal and is converted back via
+    // GetDouble() during the addition.
+    JsonUtilCode rc = dom_set_value(nullptr, doc1, ".foo", "1.5", false, false);
+    EXPECT_EQ(rc, JSONUTIL_SUCCESS);
+    rc = dom_increment_by(doc1, ".foo", &parser.Parse("1e-308", 6).GetJValue(), res, isV2Path);
+    EXPECT_EQ(rc, JSONUTIL_SUCCESS);
+    EXPECT_EQ(res.size(), 1);
+    EXPECT_EQ(res[0], 1.5 + 1e-308);
+    EXPECT_FALSE(isV2Path);
```
roshkhatri, line 797: I think here we should be testing subnormal+subnormal to see that the conversion is happening  something like 1e-308 + 2e-308. Now the answer ≈3e-308, so now the answer genuinely depends on both numbers being converted correctly.

This test doesn't actually prove the tiny number was handled correctly. It only proves "it didn't crash."

WDYT?

roshkhatri (commented):
The fix looks great, just a comment about the test case

also can we add a test for a huge negative number, which would produce negative infinity.

tst/unit/selector_test.cc
```diff
@@ -1289,6 +1289,25 @@ TEST_F(SelectorTest, test_delete) {
...
+    JsonUtilCode rc = dom_parse(nullptr, input, strlen(input), &d1);
+    EXPECT_EQ(rc, JSONUTIL_SUCCESS);
+
+    // Build the plain-digit literal 10^309, which is > DBL_MAX.
+    std::string big_literal = "1" + std::string(309, '0');
+    std::string path = "$[?(@.n > " + big_literal + ")]";
+
+    Selector selector;
+    rc = selector.getValues(*d1, path.c_str());
+    // 5 is not greater than 10^309, so no values match and it doesnt crash.
```
roshkhatri:
```suggestion
    // 5 is not greater than 10^309, so no values match and it doesn't crash.
```

roshkhatri (approved):
I think this looks good, the tests we have are the failure cases in the issue so that okay

Issue #107, opened 2026-06-05
Unbounded array Reserve in legacy encver-0 RDB load: crafted RESTORE/replication causes OOM (SIGKILL)


madolson:
### Description

The legacy node-by-node RDB load path (encoding version 0) reads an attacker-controlled array element count and passes it straight to `rapidjson`'s `Reserve()` with no bound, before reading any element. A tiny crafted `RESTORE` payload can request a multi-GB to ~64GB allocation, exhausting memory and getting the server OOM-killed (SIGKILL). Reachable via `RESTORE` and via replication, since the RDB load callback accepts encver 0 and the encoding version comes from the client-supplied payload.

### Steps to reproduce

Build a `RDB_TYPE_MODULE_2` payload with the JSON module type id's encoding-version bits (low 10 bits) cleared to 0, body = `OPCODE_UINT(JSON_METACODE_ARRAY=0x40)` + `OPCODE_UINT(<huge length>)` + `OPCODE_EOF`, then append the RDB version and a valid CRC64. `RESTORE` it.

Confirmed crashing lengths: `2^32-1`, `2^48`, `2^60`, `2^64-1`. Safe (graceful `-ERR Bad data format`): small counts like `2^24` (allocates then hits EOF cleanly).

Tooling note: CRC64 + RESTORE payload helpers used for the repro are the same ones in `bloom_rdb.py` from the bloom audit work.

**WARNING:** on an uncapped host this requests tens of GB and can take the whole machine offline via host-level OOM. Test inside a memory-capped container: `docker run --memory=512m --memory-swap=512m ...`.

### Crash output

```
# no VALKEY BUG REPORT (external SIGKILL)
docker inspect ... ExitCode=137
dmesg: Out of memory: Killed process N (valkey-server) total-vm:~69044012kB ... global_oom
```

### Root cause

`rdbLoadJValue()` in `src/json/dom.cc` (around lines 1499-1502), `JSON_METACODE_ARRAY` case:

```cpp
uint64_t length = ValkeyModule_LoadUnsigned(rdb);
array.SetArray();
array.Reserve(length, allocator);          // attacker-controlled, no bound
while (length--) array.PushBack(rdbLoadJValue(params), allocator);
```

[...]

PR #111, opened 2026-06-09
Fix unbounded array Reserve in legacy RDB load

roshkhatri:
Fix for: #107

The legacy (encoding version 0) RDB load path read an attacker-controlled array element
count and passed it straight to rapidjson's `Reserve()` with no bound, so a crafted
RESTORE or replication payload could request a ~64GB allocation and OOM-kill the server.
This fix caps the pre-reservation at `JSON_RDB_ARRAY_RESERVE_CAP` (65536) using `std::min`.
Genuine elements beyond the cap still grow the array via `PushBack` as they are actually
read, so there is no amplification from a tiny count field. The load loop now also breaks on
IO/status errors and propagates `JSONUTIL_INVALID_RDB_FORMAT`, and a check at the `dom_load`
chokepoint rejects any truncated load (object member, scalar) at the module boundary instead
of accepting a partial document.

Tests: added cases in test_rdb.py. A huge-count array payload is now rejected gracefully and
the server stays alive (without this change it balloons past 11GB), a valid legacy array still
round-trips, and a truncated array element is handled without crashing.

### DIFF
--- a/src/json/dom.cc
+++ b/src/json/dom.cc
@@ -1347,6 +1347,10 @@ enum meta_codes {
     JSON_METACODE_PAIR    = 0x80    // Codes an object Memory, a string(member name) and a JValue
 };
 
+// Cap pre-reservation for legacy RDB array loads: the declared count is untrusted
+// (RESTORE/replication), so limit upfront allocation; PushBack grows as elements arrive.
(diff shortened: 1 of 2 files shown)


zackcam (approved):
Changes and tests look good to me!

Issue #106, opened 2026-06-05
JSON.MSET with a duplicate key crashes server (SIGABRT) via failed assertion (json.cc:776)


madolson:
### Description

`JSON.MSET` that references the same key twice in one command can crash the server with a SIGABRT via a failed assertion. The command validates every key/path/value triple up front against the current stored document, then applies them sequentially. When a key appears twice, an earlier op mutates the document shape so a later op's path (which validated against the pre-mutation shape) no longer applies. The apply phase then hits `ValkeyModule_Assert(rc == JSONUTIL_SUCCESS)` and aborts. This is a TOCTOU between the validate and apply phases within a single command.

### Steps to reproduce

```
JSON.SET dk '$' '{"a":1}'
JSON.MSET dk '$' '[1,2,3]' dk '.a' '99'
```

The server aborts; the client sees `Error: Server closed the connection`.

Nested-path variant (also crashes):

```
JSON.SET dk '$' '{"a":{"x":1}}'
JSON.MSET dk '.a' '[9]' dk '.a.x' '5'
```

The key must pre-exist (a relative-path `JSON.MSET` on a missing key returns a graceful syntax error).

### Crash output

```
=== VALKEY BUG REPORT START ===
# === ASSERTION FAILED ===
# ==> src/json/json.cc:776 'rc == JSONUTIL_SUCCESS' is not true
...
_serverAssert at src/json/json.cc:776
```

### Root cause

`parseAndValidateMSetCmdArgs` (`src/json/json.cc:288`) validates each triple against the current stored doc (`dom_verify_value` -> `selector.prepareSetValues` on `doc->GetJValue()`). The apply loop in `Command_JsonMSet` (`src/json/json.cc:752-786`) mutates keys sequentially. For a duplicate key, op1 (e.g. root replace `$ [1,2,3]`) changes the document shape, so op2's path (`.a`), which validated against the original object, no longer resolves. `dom_set_value` (`src/json/json.cc:775`) returns a non-SUCCESS code and the guard `ValkeyModule_Assert(rc == JSONUTIL_SUCCESS)` at `src/json/json.cc:776` aborts the process. Validate and apply phases operate on different document states for the same key.

### Suggested fix

Re-validate each op against the live document inside the apply loop and convert failures to `ReplyWithError`, or reject duplicate keys in `JSON.MSET`.

### Environment

- valkey-json commit `a932a1c`, `valkey/valkey-bundle:unstable`
- Reproduced on Linux x86_64

---

[...]

PR #112, opened 2026-06-10
Fix JSON.MSET duplicate-key crash

roshkhatri:
Fix for: #106

A `JSON.MSET` that references the same key more than once could crash the server
(SIGABRT). Triples are validated up front and then applied in place under
`ValkeyModule_Assert(rc == JSONUTIL_SUCCESS)`. For a duplicated key, an earlier op changes
the document's shape, so a later op (validated against the original shape) fails at apply
time and the assertion aborts the process.

This removes the apply-phase assertions and applies ops on a best-effort basis. Each key
is opened fresh at apply time, so every op sees the current value, including the effect of
an earlier op on the same key. An op that cannot apply against the current document is
skipped and the remaining ops are still applied; the command returns OK. Up-front
validation is unchanged, so malformed commands are still rejected atomically with an error
before any write.

This also makes JSON.MSET behavior compatible with RedisJSON.

Tests: added cases for a duplicated key whose later op conflicts with the first (returns
OK, server alive, first op committed), a three-op case where a middle op is skipped and
the others applied, and a case where the later op is valid against the first op's result
so both apply. Existing MSET and validation tests pass unchanged; the full integration
suite passes across all server versions including the ASAN build.

### DIFF
--- a/src/json/json.cc
+++ b/src/json/json.cc
@@ -278,7 +278,6 @@ STATIC JsonUtilCode parseSetCmdArgs(ValkeyModuleString **argv, const int argc, S
 
 typedef struct {
     ValkeyModuleString *key_str;    // Required
-    ValkeyModuleKey *key;
     const char *path;               // Required
     const char *json;               // Required
     size_t json_len;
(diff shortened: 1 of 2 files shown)


zackcam (approved):
Looks good to me, should we put a note in the docs that the behaviour now will silently fail on a bad duplicate key but still return ok?

PR #114, opened 2026-06-22
Making it so ASAN failures fail on CI, fixing heap-use-after-free bugs in JSON.MSET by closing key handles (ValkeyModule_CloseKey) in all code paths within the operation loop

zackcam:
JSON.MSET opened key handles in both the parse/validation loop and the apply loop but never closed them, relying on AutoMemory to clean up at command end.

For duplicate keys, a later op's ValkeyModule_ModuleTypeSetValue does a VM_DeleteKey + re-add, freeing the key object that an earlier still-open handle points at. AutoMemory then closes that stale handle at command teardown -> heap-use-after-free (ASAN: moduleCloseKey, module.c:4456).

Fix: ValkeyModule_CloseKey each handle as soon as it is no longer needed (after validation in the parse loop, and at the end of every apply-loop iteration including the skip/continue paths), so at most one live handle exists per key.

Reproduced and verified under ASAN (amztests-asan) on the exact failing tests. This also exposed that ASAN failures were not being caught properly in the CI runs. This PR also fixes this but still allows the full suite of tests to run and check for failures and show as a fail at the end.

Example run of failing: https://github.com/zackcam/valkey-json/actions/runs/27984788018/job/82823547063 this was a run with the new CI changes but missing the MSET fixes and we can see it failed on ASAN.

### DIFF
--- a/src/json/json.cc
+++ b/src/json/json.cc
@@ -368,6 +368,8 @@ STATIC JsonUtilCode parseAndValidateMSetCmdArgs(ValkeyModuleCtx *ctx, ValkeyModu
                 return rc;
             }
         }
+        // Close the validation handle so no stale handle to a duplicate key survives.
+        ValkeyModule_CloseKey(key);
     }
     return JSONUTIL_SUCCESS;
 }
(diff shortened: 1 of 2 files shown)


zackcam:
Closing to make two different prs

PR #117, opened 2026-07-29
Making PR's run unstable and latest only, adding a daily job for all versions

zackcam:
This pr has set up a daily job to run JSON against all versions of valkey (If this becomes unwieldily we can easily amend to limit number of versions). In addition it now will run pr requests only against unstable and the latest version which is got dynamically so will not need to be updated on releases. The build script also now contains a dynamic way of checking and asserting what versions are availble

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,14 +1,48 @@
 name: CI
 
-on: [push, pull_request]
+on:
+  push:
+  pull_request:
+  # All versions run at 06:00 UTC every day (full backwards-compat matrix).
+  schedule:
+    - cron: '0 6 * * *'
+  workflow_dispatch:
 
 jobs:
+  get-versions:
+    runs-on: ubuntu-latest
+    outputs:
+      versions: ${{ steps.discover.outputs.versions }}
+    steps:
+      - name: Discover valkey server versions
+        id: discover
+        env:
+          VALKEY_REPO_URL: https://github.com/valkey-io/valkey.git
+          MIN_SERVER_MAJOR: '8'
+        run: |
+          branches=$(git ls-remote --heads "$VALKEY_REPO_URL" \
+            | sed 's#.*refs/heads/##' \
+            | grep -E '^[0-9]+\.[0-9]+$' \
+            | awk -F. -v M="$MIN_SERVER_MAJOR" '$1>=M' \
+            | sort -V)
+          # PRs/pushes: fast signal — only "unstable" + the latest official release.
+          # schedule/workflow_dispatch: full supported matrix for backwards-compat.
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -1,14 +1,48 @@
...
+            | awk -F. -v M="$MIN_SERVER_MAJOR" '$1>=M' \
+            | sort -V)
+          # PRs/pushes: fast signal — only "unstable" + the latest official release.
+          # schedule/workflow_dispatch: full supported matrix for backwards-compat.
+          if [ "$GITHUB_EVENT_NAME" = "schedule" ] || [ "$GITHUB_EVENT_NAME" = "workflow_dispatch" ]; then
+            selected="$branches"
+          else
+            selected=$(printf '%s\n' $branches | tail -n 1)
+          fi
+          json=$(printf '%s\n' unstable $selected | jq -R . | jq -sc .)
```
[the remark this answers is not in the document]
roshkhatri: If there is a network issue, we wont be able to pull too, so this is not relevant here

.github/workflows/ci.yml
```diff
@@ -1,14 +1,48 @@
 name: CI
 
-on: [push, pull_request]
+on:
+  push:
+  pull_request:
+  # All versions run at 06:00 UTC every day (full backwards-compat matrix).
+  schedule:
+    - cron: '0 6 * * *'
```
roshkhatri, line 8: This is a good addition, would work like daily/weekly in valkey

roshkhatri (dismissed):
LGTM

build.sh
```diff
@@ -79,6 +79,18 @@ if [ -z "$SERVER_VERSION" ]; then
...
+VALID_VERSIONS=$(printf 'unstable\n%s\n' "$(git ls-remote --heads "$REPO_URL" \
+  | sed 's#.*refs/heads/##' \
+  | grep -E '^[0-9]+\.[0-9]+$' \
+  | awk -F. -v M="$MIN_SERVER_MAJOR" '$1>=M')")
+if ! printf '%s\n' "$VALID_VERSIONS" | grep -qxF "$SERVER_VERSION"; then
+  echo "ERROR: Unsupported version - $SERVER_VERSION"
+  echo "Valid versions are: $(printf '%s\n' "$VALID_VERSIONS" | sort -V | tr '\n' ' ')"
+  exit 1
+fi
+
```
roshkhatri, line 100: On another thought this will break the offline build. might want to add a condition for when this is called

zackcam: Just fixed, made it so we log but continue. We will still fail if no binary is there but that was the case before

roshkhatri (approved):

PR #120, opened 2026-08-31
Fixing crash on recursive path mutation

roshkhatri:
## Summary

Fixes a use-after-free crash when a recursive JSONPath expression selects both a
parent array and a descendant nested inside it, during a size-changing mutation.
Affects `JSON.ARRAPPEND`, `JSON.ARRINSERT`, `JSON.ARRPOP`, `JSON.ARRTRIM` and
`JSON.CLEAR`.

A node's `JValue*` points into its parent array's element buffer. When the parent
is mutated, that buffer is reallocated or freed, leaving the descendant's cached
pointer dangling; dereferencing it is a use-after-free.

## Fix

- Add `order_result_set_deepest_first()`, which stable-sorts matches by path depth
  so descendants are processed before their ancestors. A parent's buffer is never
  reallocated before its children have been consumed.
- Refactor the four array mutators to write results by index into a pre-sized
  vector, so replies stay in document order while processing runs depth-first.
  `internal_array_pop`, `internal_array_insert` and `internal_array_trim` now take
  a single out-parameter instead of appending to the result vector.
- `dom_clear` only returns a count, so ordering there is irrelevant.

For result sets with no parent/descendant overlap the sort is a stable no-op, so
existing behaviour and reply ordering are unchanged.

### DIFF
--- a/src/json/dom.cc
+++ b/src/json/dom.cc
@@ -8,6 +8,7 @@
 #include <iomanip>
 #include <cmath>
 #include <string>
+#include <algorithm>
 #include "json/rapidjson_includes.h"
 #include <rapidjson/pointer.h>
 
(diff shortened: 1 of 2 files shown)


madolson (approved):
Looks fine, but Cameron should confirm.

zackcam (approved):

roshkhatri:
Thanks Cameron!

PR #124, opened 2026-09-17
Fix overflow in JSONPath slice iteration

roshkhatri:
## Summary

- use a 64-bit iterator for JSONPath slice evaluation so large steps cannot be truncated
- stop iteration before extreme positive or negative steps can overflow the iterator
- parse the full signed 64-bit range safely, including `INT64_MIN`, and reject out-of-range integer literals
- add unit and integration regression coverage for steps of ±2^32, signed 64-bit limits, and invalid values such as 2^64

## Testing

- focused `DomTest.testSelector_get_array_slice_large_step` unit test
- focused `TestJsonBasic.test_json_get_v2path_array_slice` integration test against Valkey unstable


roshkhatri:
Addressed the review finding in f9707ee: integer parsing now handles `INT64_MIN` without signed overflow and rejects values outside the signed 64-bit range, including `2^64`. Added unit and integration regression coverage for the accepted and rejected boundaries.

